PR #13, opened 2021-05-18
Improve integration with data streams

### LINKED ISSUE #12: [Feature] Improve integration with data streams
**Context**

A data stream is an index abstraction very similar to an index alias. Under the hood, it points to multiple backing indices and allows searches using a single named resource. The underlying backing indices of a data stream are automatically created and follow this naming convention: `.ds-<data-stream-name>-<generation>`

For example, a data stream named `logs-redis` will have backing indices named `.ds-logs-redis-000001`, `.ds-logs-redis-000002`, and so on.

The creation of a data stream requires a matching index template containing the mappings and settings used to configure the data stream's backing indices. The `data_stream` field indicates that the template creates a data stream instead of a regular index.

```
PUT /_index_template/my-data-stream-template
{
    "index_patterns": [ "logs-haproxy", "logs-nginx", "logs-redis" ],
    "data_stream": { }
}
```

Note that the `index_patterns` in the template matches the data stream name, not the underlying backing indices.

**Is your feature request related to a problem? Please describe.**

We have identified some limitations/enhancements when using the Index Management plugin with data streams.
1. Unable to associate an ISM policy with a data stream
    To automatically associate an ISM policy with newly created indices, users can define index patterns within the policy itself. While it is possible to define an index pattern that matches the name of the underlying backing indices of a data stream (eg. `.ds-logs-redis-*`), it is not yet possible to define an index pattern that matches the name of the data stream directly (eg. `logs-redis`). This differs from how index patterns are matched in the case of an index template.

    Suggestion:
    When a newly created index belongs to a data stream, the ISM policy of the highest priority having an index pattern matching the parent data stream name should be associated with the index. This will make the index pattern matching behavior consistent with index templates.

2. Rollover alias should not be needed for a data stream
    ISM supports the rollover action but is primarily built for index aliases as the rollover target. For this reason, ISM expects the `rollover_alias` index setting to be defined in the index template. This isn't necessary for data streams because:
    + Data streams do not use an index alias, so the name `rollover_alias` is confusing.
    + ISM should be able to identify the data stream name (i.e. the rollover target) from the backing index itself.
    + Multiple data streams can be created from a single index template, but the `rollover_alias` setting in that template can only have one rollover target.
    
    Suggestion:
    When executing the rollover action on a backing index of a data stream, ISM should use the data stream name as the rollover target. Users will no longer have to define the `rollover_alias` setting for a data stream.

3. Rollup jobs do not resolve the backing indices of a data stream
    A rollup job allows users to define index expressions as the source. These index expressions can resolve index names, index patterns, and index aliases to the list of corresponding concrete indices. The same resolution doesn't happen for data streams.
    
    Suggestion:
    Index expression resolution should resolve a data stream name to its corresponding backing indices. This will make the resolution behavior consistent with index aliases.

**Additional context**

This change builds on top of the following feature to support data streams in OpenSearch:
https://github.com/opensearch-project/OpenSearch/issues/675

ketanv3:
### Description

This PR aims to resolve the limitations/enhancements pointed out in https://github.com/opensearch-project/index-management/issues/12. Here's a short summary.

1. Fixes the ISM policy association behaviour when a new backing index of a data stream is created.
2. When executing the rollover action on a backing index of a data stream, ISM will automatically identify the data stream name and use it as the rollover target. Users no longer need to specify the `rollover_alias` index setting.
3. Fixes the index expression resolution in rollups when the source is a data stream. The data stream name will be resolved to its corresponding backing indices.

### Issues
+ Closes https://github.com/opensearch-project/index-management/issues/12
+ Relates https://github.com/opensearch-project/OpenSearch/issues/675

### CheckList
- [x] New functionality includes tests
- [x] All tests pass

Signed-off-by: Ketan Verma ketan9495@gmail.com

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
@@ -31,6 +31,7 @@ import org.opensearch.indexmanagement.util.IndexManagementException
 import org.apache.logging.log4j.LogManager
 import org.apache.lucene.util.automaton.Operations
 import org.opensearch.OpenSearchException
+import org.opensearch.cluster.ClusterState
 import org.opensearch.cluster.metadata.IndexMetadata
 import org.opensearch.common.Strings
 import org.opensearch.common.ValidationException
@@ -49,18 +50,26 @@ private val log = LogManager.getLogger("ISMTemplateService")
  * @return policyID
  */
 @Suppress("ReturnCount")
-fun Map<String, ISMTemplate>.findMatchingPolicy(indexMetadata: IndexMetadata): String? {
+fun Map<String, ISMTemplate>.findMatchingPolicy(clusterState: ClusterState, indexName: String): String? {
     if (this.isEmpty()) return null
 
-    val indexName = indexMetadata.index.name
+    val indexMetadata = clusterState.metadata.index(indexName)
+    val indexAbstraction = clusterState.metadata.indicesLookup[indexName]!!
+    val isDataStreamIndex = indexAbstraction.parentDataStream != null
 
-    // don't include hidden index
+    // Don't include hidden index unless it belongs to a data stream.
     val isHidden = IndexMetadata.INDEX_HIDDEN_SETTING.get(indexMetadata.settings)
-    if (isHidden) return null
+    if (!isDataStreamIndex && isHidden) return null
+
+    // If the index belongs to a data stream, then find the matching policy using the data stream name.
+    val lookupName = when {
+        isDataStreamIndex -> indexAbstraction.parentDataStream!!.name!!
+        else -> indexName
+    }
 
     // only process indices created after template
     // traverse all ism templates for matching ones
-    val patternMatchPredicate = { pattern: String -> Regex.simpleMatch(pattern, indexName) }
+    val patternMatchPredicate = { pattern: String -> Regex.simpleMatch(pattern, lookupName) }
     var matchedPolicy: String? = null
     var highestPriority: Int = -1
     this.filter { (_, template) ->
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
@@ -54,8 +54,8 @@ fun Map<String, ISMTemplate>.findMatchingPolicy(clusterState: ClusterState, inde
     if (this.isEmpty()) return null
 
     val indexMetadata = clusterState.metadata.index(indexName)
-    val indexAbstraction = clusterState.metadata.indicesLookup[indexName]!!
-    val isDataStreamIndex = indexAbstraction.parentDataStream != null
+    val indexAbstraction = clusterState.metadata.indicesLookup[indexName]
+    val isDataStreamIndex = indexAbstraction?.parentDataStream != null
 
     // Don't include hidden index unless it belongs to a data stream.
     val isHidden = IndexMetadata.INDEX_HIDDEN_SETTING.get(indexMetadata.settings)
@@ -63,7 +63,7 @@ fun Map<String, ISMTemplate>.findMatchingPolicy(clusterState: ClusterState, inde
 
     // If the index belongs to a data stream, then find the matching policy using the data stream name.
     val lookupName = when {
-        isDataStreamIndex -> indexAbstraction.parentDataStream!!.name!!
+        isDataStreamIndex -> indexAbstraction?.parentDataStream?.name
         else -> indexName
     }
 
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
@@ -137,14 +137,19 @@ class AttemptRolloverStep(
             // 1. IndexAbstraction.Type.DATA_STREAM - the new index is added to the data stream indicated by the 'rolloverTarget'
             // 2. IndexAbstraction.Type.ALIAS - the new index is added to the alias indicated by the 'rolloverTarget'
             if (response.isAcknowledged) {
+                val message = when {
+                    isDataStream -> getSuccessDataStreamRolloverMessage(rolloverTarget, indexName)
+                    else -> getSuccessMessage(indexName)
+                }
+
                 stepStatus = StepStatus.COMPLETED
                 info = listOfNotNull(
-                    "message" to getSuccessMessage(indexName),
+                    "message" to message,
                     if (conditions.isEmpty()) null else "conditions" to conditions // don't show empty conditions object if no conditions specified
                 ).toMap()
             } else {
                 val message = when {
-                    isDataStream -> getFailedDataStreamUpdateMessage(rolloverTarget)
+                    isDataStream -> getFailedDataStreamRolloverMessage(rolloverTarget)
 
                     // If the alias update response was NOT acknowledged, then the new index was created but we failed to swap the alias
                     else -> getFailedAliasUpdateMessage(indexName, response.newIndex)
@@ -165,11 +170,11 @@ class AttemptRolloverStep(
 
     private fun getRolloverTargetOrUpdateInfo(): Pair<String?, Boolean> {
         val metadata = clusterService.state().metadata()
-        val indexAbstraction = metadata.indicesLookup[indexName]!!
-        val isDataStreamIndex = indexAbstraction.parentDataStream != null
+        val indexAbstraction = metadata.indicesLookup[indexName]
+        val isDataStreamIndex = indexAbstraction?.parentDataStream != null
 
         val rolloverTarget = when {
-            isDataStreamIndex -> indexAbstraction.parentDataStream!!.name!!
+            isDataStreamIndex -> indexAbstraction?.parentDataStream?.name
             else -> metadata.index(indexName).getRolloverAlias()
         }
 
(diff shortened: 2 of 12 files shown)


ketanv3:
Please note that the newly added integration tests are expected to fail.

Changes to add data streams' support in OpenSearch have been merged recently (https://github.com/opensearch-project/OpenSearch/pull/690), but the latest artifacts are yet to be made available.

The test clusters launched from the existing artifact (`opensearch-1.0.0-beta1-linux-x64.tar.gz`) do not support the creation of a data stream, thus, newly added integration tests that involve the creation of a data stream are expected to fail. Will have to re-run the checks once the latest artifacts are available.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -219,6 +231,7 @@ class AttemptRolloverStep(
         fun getFailedMessage(index: String) = "Failed to rollover index [index=$index]"
         fun getFailedAliasUpdateMessage(index: String, newIndex: String) =
             "New index created, but failed to update alias [index=$index, newIndex=$newIndex]"
+        fun getFailedDataStreamUpdateMessage(dataStream: String) = "Failed to rollover data stream [data_stream=$dataStream]"
```
bowenlan-amzn, line 234: 1. np: can we change this to `getFailedDataStreamRolloverMessage`

2. can we also have a `getSuccessDataStreamRolloverMessage`, because currently we only have `getSuccessMessage `

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -125,25 +125,30 @@ class AttemptRolloverStep(
...
             // Do not need to check for isRolledOver as we are not passing any conditions or dryrun=true
             // which are the only two ways it comes back as false
 
-            // If response isAcknowledged it means the index was created and alias was added to new index
+            // If the response is acknowledged, then the new index is created and added to one of the following index abstractions:
+            // 1. IndexAbstraction.Type.DATA_STREAM - the new index is added to the data stream indicated by the 'rolloverTarget'
+            // 2. IndexAbstraction.Type.ALIAS - the new index is added to the alias indicated by the 'rolloverTarget'
             if (response.isAcknowledged) {
                 stepStatus = StepStatus.COMPLETED
                 info = listOfNotNull(
```
bowenlan-amzn, line 141: can we also distinguish datastream and index here, like below dealing with the Failed situation. Giving datastream success a different message here.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -158,17 +163,24 @@ class AttemptRolloverStep(
         }
     }
 
-    private fun getAliasOrUpdateInfo(): String? {
-        val alias = clusterService.state().metadata().index(indexName).getRolloverAlias()
+    private fun getRolloverTargetOrUpdateInfo(): Pair<String?, Boolean> {
+        val metadata = clusterService.state().metadata()
+        val indexAbstraction = metadata.indicesLookup[indexName]!!
```
bowenlan-amzn, line 168: sorry we dont allow `!!` to be in the source code. Can we pls refactor this to sth like:
```kotlin
val indexAbstraction = metadata.indicesLookup[indexName]
if (indexAbstraction != null)
...
```

same for other `!!` usecase here

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
```diff
@@ -49,18 +50,26 @@ private val log = LogManager.getLogger("ISMTemplateService")
  * @return policyID
  */
 @Suppress("ReturnCount")
-fun Map<String, ISMTemplate>.findMatchingPolicy(indexMetadata: IndexMetadata): String? {
+fun Map<String, ISMTemplate>.findMatchingPolicy(clusterState: ClusterState, indexName: String): String? {
     if (this.isEmpty()) return null
 
-    val indexName = indexMetadata.index.name
+    val indexMetadata = clusterState.metadata.index(indexName)
+    val indexAbstraction = clusterState.metadata.indicesLookup[indexName]!!
```
bowenlan-amzn, line 57: sorry we dont allow `!!` to be used in src code, it is considered as code smell. Can we refactor this to explicity check null.

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/RolloverActionIT.kt
```diff
@@ -264,4 +266,143 @@ class RolloverActionIT : IndexStateManagementRestTestCase() {
...
+            ismTemplate = ISMTemplate(listOf(dataStreamName), 100, Instant.now().truncatedTo(ChronoUnit.MILLIS))
+        )
+        createPolicy(policy, policyID)
+
+        // Create the data stream
+        val firstIndexName = DataStream.getDefaultBackingIndexName(dataStreamName, 1L)
+        client().makeRequest(
+            "PUT",
+            "/_index_template/rollover-data-stream-template",
+            StringEntity("{ \"index_patterns\": [ \"$dataStreamName\" ], \"data_stream\": { } }", ContentType.APPLICATION_JSON)
```
bowenlan-amzn, line 294: Curious about `index_patterns` format here. Take `logs-redis` for an example, normally we will use `logs-redis*` as the index pattern with `*` in the end. But it seems here we use `logs-redis`, is this sth by design? And `logs-redis*` is wrong?

src/main/kotlin/org/opensearch/indexmanagement/rollup/actionfilter/FieldCapsFilter.kt
```diff
@@ -84,7 +84,7 @@ class FieldCapsFilter(
             val localIndices = remoteClusterIndices.remove(RemoteClusterAware.LOCAL_CLUSTER_GROUP_KEY)
 
             localIndices?.let {
-                val concreteIndices = indexNameExpressionResolver.concreteIndexNames(clusterService.state(), request.indicesOptions(), it)
+                val concreteIndices = indexNameExpressionResolver.concreteIndexNames(clusterService.state(), it)
```
thalurur, line 87: not sure if/why we want to do this here? Rollup target index cannot be a data stream index

bowenlan-amzn:
Pls help update our release note for rc1 before merge in: https://github.com/opensearch-project/index-management/blob/main/release-notes/opensearch-index-management.release-notes-1.0.0.0-rc1.md

Add this part to the release note
```
### Enhancements

* Improve integration with data streams ([#13](https://github.com/opensearch-project/index-management/pull/13))
```

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
```diff
@@ -101,9 +101,7 @@ class RollupInterceptor(
                             throw IllegalArgumentException("Rollup search must have size explicitly set to 0, but found ${request.source().size()}")
                         }
 
-                        val indices = request.indices().map { it.toString() }.toTypedArray()
-                        val concreteIndices = indexNameExpressionResolver
-                                .concreteIndexNames(clusterService.state(), request.indicesOptions(), *indices)
+                        val concreteIndices = indexNameExpressionResolver.concreteIndexNames(clusterService.state(), request)
```
thalurur, line 104: target indices cannot be data streams so not sure if this need to be changed

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/RolloverActionIT.kt
```diff
@@ -264,4 +266,143 @@ class RolloverActionIT : IndexStateManagementRestTestCase() {
...
+        // Speed up to the second execution of the policy where it will trigger the first execution of the action.
+        updateManagedIndexConfigStartTime(managedIndexConfig)
+        waitFor {
+            val info = getExplainManagedIndexMetaData(firstIndexName).info as Map<String, Any?>
+            assertEquals("Index did not rollover.", AttemptRolloverStep.getSuccessMessage(firstIndexName), info["message"])
+            assertNull("Should not have conditions if none specified", info["conditions"])
+        }
+
+        val secondIndexName = DataStream.getDefaultBackingIndexName(dataStreamName, 2L)
+        Assert.assertTrue("New rollover index does not exist.", indexExists(secondIndexName))
```
bowenlan-amzn, line 313: Can we add more lines under this to check if this second index is auto managed by ISM.

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/RolloverActionIT.kt
ketanv3: The index pattern matches the name of the data stream instead of the backing indices. This is by design and consistent with how index patterns work in index templates as well.

A data stream named `logs-redis` will have backing indices like `.ds-logs-redis-000001`, `.ds-logs-redis-000002` and so on. When defining an index pattern like `index_pattern: ["logs-redis"]`, I'm just performing an exact match on one data stream.

I could also define `index_pattern: ["logs-redis-*"]` which would allow me to match multiple data streams (and effectively their backing indices). This would match `"logs-redis-prod"`, `"logs-redis-staging"` and so on.

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
ketanv3: We were looking at a future enhancement where a rollup index can be added back to the same data stream. Given that querying a raw and rollup index isn't supported yet, I'll remove this change and track it in a separate issue/PR.

src/main/kotlin/org/opensearch/indexmanagement/rollup/actionfilter/FieldCapsFilter.kt
ketanv3: This is equivalent to the previous statement. Will remove it to avoid the diff.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
ketanv3: Updated these.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
ketanv3: Added a different success message for data streams.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
ketanv3: Thanks for pointing this out.
I have replaced these with Kotlin safe-calls `?.` to remove the code smell while keeping the code succinct.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
ketanv3: I have replaced these with Kotlin safe-calls `?.` to remove the code smell while keeping the code succinct.

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/RolloverActionIT.kt
ketanv3: Added this check.

bowenlan-amzn:
We have looked into the one failure in multi-node tests. `test change policy API should only apply to indices in the state filter`

This is a flakey one, doesn't seem to be related to changes in this PR.

bowenlan-amzn (approved):

PR #14, opened 2021-05-21
Renaming Namespaces

bowenlan-amzn:
*Issue #, if available:*
#11 

*Description of changes:*

Rename Namespaces (e.g. com.amazon.opendistroforelasticsearch to org.opensearch)

### Test

Running test against a local OpenSearch 1.x cluster, tests failures are expected.

Github workflow can be updated when we have a new tarball release

*CheckList:*
[ ] Commits are signed per the DCO using --signoff


.github/workflows/multi-node-test-workflow.yml
```diff
@@ -28,40 +28,40 @@ jobs:
         with:
           repository: 'opensearch-project/OpenSearch'
           path: OpenSearch
-          ref: '1.0.0-beta1'
+          ref: '1.x'
```
lezzago, line 31: Why not `1.0.0-rc1` instead?
There is a tag cut for it: https://github.com/opensearch-project/OpenSearch/tree/1.0.0-rc1

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
```diff
@@ -83,7 +83,7 @@ class AttemptSnapshotStep(
                     .format(DateTimeFormatter.ofPattern("uuuu.MM.dd-HH:mm:ss.SSS", Locale.ROOT)))
 
             val createSnapshotRequest = CreateSnapshotRequest()
-                .userMetadata(mapOf("snapshot_created" to "Open Distro for Elasticsearch Index Management"))
+                .userMetadata(mapOf("snapshot_created" to "OpenSearch Index Management"))
```
thalurur, line 86: Are you sure we want to change this? this can potentially break customer filters if they are filtering  snapshots based on this string today?

.github/workflows/multi-node-test-workflow.yml
bowenlan-amzn: should be fine, I think it's same for this release. As long as this can help us run test successfully

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
bowenlan-amzn: revert back, since there is concern it could break customer existing script.

thalurur (approved):

lezzago (approved):

PR #22, opened 2021-06-01
Adding more checks around schedule and metadata in transforms

thalurur:

*Issue #, if available:*

*Description of changes:*
* Adding additional check when creating transform to check period is set to be greater than 0
* Not allowing setting metadata id for new transform 

*CheckList:*
[ X] Commits are signed per the DCO using --signoff

### DIFF
--- a/src/test/kotlin/org/opensearch/indexmanagement/transform/model/XContentTests.kt
+++ b/src/test/kotlin/org/opensearch/indexmanagement/transform/model/XContentTests.kt
@@ -11,6 +11,7 @@
 
 package org.opensearch.indexmanagement.transform.model
 
+import org.junit.Assert
 import org.opensearch.indexmanagement.opensearchapi.parseWithType
 import org.opensearch.indexmanagement.indexstatemanagement.util.XCONTENT_WITHOUT_TYPE
 import org.opensearch.indexmanagement.transform.randomTransform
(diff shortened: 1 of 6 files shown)


src/test/kotlin/org/opensearch/indexmanagement/transform/model/XContentTests.kt
```diff
@@ -58,6 +58,13 @@ class XContentTests : OpenSearchTestCase() {
         assertEquals("Round tripping Transform with type doesn't work", transform, parsedTransform)
     }
 
+    fun `test transform parsing should ignore metadata id and startTime if its newly created transform`() {
+        val transform = randomTransform().copy(metadataId = "randomMetadata")
```
bowenlan-amzn, line 62: can user manually change metadataId? if so, when do they need to do that.

thalurur: initially the job is created with no metadata, and later its is updated to have a metadata. This is the only valid case where metadata id is updated, there is no other purpose for customer to update the metadata

thalurur: the update API is the only way to set metadata id, this API is an external API meaning users can call it to set any metadata id but if they do it will potentially break the job

bowenlan-amzn (approved):

PR #77, opened 2021-06-10
Making snapshot name to scripted input in template 

thalurur:

*Issue #, if available:*
https://discuss.opendistrocommunity.dev/t/universal-snapshot-policy-for-all-indexes/4815

*Description of changes:*
Processing snapshot field as a script instead of string so snapshot name can be derived from run time context - the two attributes supported during runtime are index and indexuuid

*CheckList:*
[X] Commits are signed per the DCO using --signoff

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
@@ -80,13 +80,13 @@ class AttemptSnapshotStep(
                 info = mutableInfo.toMap()
                 return this
             }
-            val snapshotNamePrefix = "-".plus(
+            val snapshotNameSuffix = "-".plus(
                 LocalDateTime.now(ZoneId.of("UTC"))
                     .format(DateTimeFormatter.ofPattern("uuuu.MM.dd-HH:mm:ss.SSS", Locale.ROOT))
             )
 
             val snapshotScript = Script(ScriptType.INLINE, Script.DEFAULT_TEMPLATE_LANG, config.snapshot, mapOf())
-            snapshotName = compileTemplate(snapshotScript, managedIndexMetaData, indexName).plus(snapshotNamePrefix)
+            snapshotName = compileTemplate(snapshotScript, managedIndexMetaData, indexName).plus(snapshotNameSuffix)
 
             val createSnapshotRequest = CreateSnapshotRequest()
                 .userMetadata(mapOf("snapshot_created" to "Open Distro for Elasticsearch Index Management"))
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
@@ -86,7 +86,9 @@ class AttemptSnapshotStep(
             )
 
             val snapshotScript = Script(ScriptType.INLINE, Script.DEFAULT_TEMPLATE_LANG, config.snapshot, mapOf())
-            snapshotName = compileTemplate(snapshotScript, managedIndexMetaData, indexName).plus(snapshotNameSuffix)
+            // If user intentionally set the snapshot name empty then we are going to honor it
+            val defaultSnapshotName = if (config.snapshot.isBlank()) config.snapshot else indexName
+            snapshotName = compileTemplate(snapshotScript, managedIndexMetaData, defaultSnapshotName).plus(snapshotNameSuffix)
 
             val createSnapshotRequest = CreateSnapshotRequest()
                 .userMetadata(mapOf("snapshot_created" to "Open Distro for Elasticsearch Index Management"))
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
@@ -44,6 +44,7 @@ import org.opensearch.indexmanagement.opensearchapi.suspendUntil
 import org.opensearch.rest.RestStatus
 import org.opensearch.script.Script
 import org.opensearch.script.ScriptService
+import org.opensearch.script.ScriptType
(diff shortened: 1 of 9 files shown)


src/test/resources/mappings/cached-opendistro-ism-config.json
```diff
@@ -194,7 +194,8 @@
                       "type": "keyword"
                     },
                     "snapshot": {
-                      "type": "keyword"
```
dbbaughe, line 197: Can't change the type of an existing mapping

thalurur: yes, I am still checking what is the best way to approach this

cryptk:
Is there something that is holding this up? This feature would be greatly appreciated

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
```diff
@@ -74,15 +80,13 @@ class AttemptSnapshotStep(
                 info = mutableInfo.toMap()
                 return this
             }
+            val snapshotNamePrefix = "-".plus(
```
dbbaughe, line 83: nit: postfix/suffix

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
```diff
@@ -159,6 +173,7 @@ class AttemptSnapshotStep(
     }
 
     companion object {
+        val validTopContextFields = setOf("index", "indexUuid")
```
dbbaughe, line 176: Just to confirm is this all users have asked for? (Access to index/indexUuid in the snapshot name)

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
```diff
@@ -74,15 +80,13 @@ class AttemptSnapshotStep(
...
-            snapshotName = config
-                .snapshot
-                .plus("-")
-                .plus(
-                    LocalDateTime
-                        .now(ZoneId.of("UTC"))
-                        .format(DateTimeFormatter.ofPattern("uuuu.MM.dd-HH:mm:ss.SSS", Locale.ROOT))
-                )
+            val snapshotScript = Script(ScriptType.INLINE, Script.DEFAULT_TEMPLATE_LANG, config.snapshot, mapOf())
+            snapshotName = compileTemplate(snapshotScript, managedIndexMetaData, indexName).plus(snapshotNamePrefix)
```
dbbaughe, line 89: Any behavior differences for users w/ existing snapshot actions or they should continue to work (and behave) as before this fix?

dbbaughe: i.e. their snapshot names should not change somehow because of this change if they are not using the ctx variable and are just using a string.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
```diff
@@ -148,6 +152,16 @@ class AttemptSnapshotStep(
...
     }
 
+    private fun compileTemplate(template: Script, managedIndexMetaData: ManagedIndexMetaData, defaultValue: String): String {
+        val contextMap = managedIndexMetaData.convertToMap().filterKeys { key ->
+            key in validTopContextFields
+        }
+        val compiledValue = scriptService.compile(template, TemplateScript.CONTEXT)
+            .newInstance(template.params + mapOf("ctx" to contextMap))
+            .execute()
+        return if (compiledValue.isBlank()) defaultValue else compiledValue
```
dbbaughe, line 162: When does compiledValue.isBlank() occur?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
thalurur: will change

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
thalurur: the users only asked for indexName in the issue, I have added indexUuid too since its readily available

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
thalurur: They should work the same way, the mustache template compiler will return the same string value if there is no placeholder fields

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/snapshot/AttemptSnapshotStep.kt
thalurur: if someone puts in a template with just a replaceable field and that is not valid e.g. ``{{ctx.name}}`` or ``{{ctx.anything_not_allowed}}`` this will be empty string

If someone specifies the template as ``{{ctx.anything_not_allowed}}_snapshot`` then the compiled value is just `` _snapshot``

dbbaughe: Hmm can someone specify an empty string as the snapshot name currently or do we block that?
i.e. if currently sameone has "" and it's allowed then instead of "" + the postfix stuff it'll change to $indexName + postfix stuff (since it looks like we add the index name as the default to use if this returns blank)

thalurur: good point, there is not check for it being non empty at the moment the field needs to be present in the policy but it can be empty string. 

I can update the default value to empty string in that case - i.e if the original name is empty string

dbbaughe (approved):

Issue #82, opened 2021-06-22
[BUG] Error messages in logs.


skkosuri-amzn:
**Describe the bug**
On OpenSearch cluster, I see below errors messages. I didnt enable ISM or using ISM in this cluster.

[2021-06-22T09:55:32,208][ERROR][o.o.i.i.ManagedIndexCoordinator] [node-1] Failed to clear ManagedIndexMetadata for index uuid: [rkYN9BFNQm-C_7Jcc0yISA#metadata], failureMessage: [.opendistro-ism-config] IndexNotFoundException[no such index [.opendistro-ism-config]]
[2021-06-22T09:55:32,208][ERROR][o.o.i.i.ManagedIndexCoordinator] [node-1] Failed to clear ManagedIndexMetadata for index uuid: [gABHBr6qRZKmrDJ72saDKA#metadata], failureMessage: [.opendistro-ism-config] IndexNotFoundException[no such index [.opendistro-ism-config]]
[2021-06-22T09:55:32,208][ERROR][o.o.i.i.ManagedIndexCoordinator] [node-1] Failed to clear ManagedIndexMetadata for index uuid: [7iTRjPayRSuc6PP-ZuZfgQ#metadata], failureMessage: [.opendistro-ism-config] IndexNotFoundException[no such index [.opendistro-ism-config]]
[2021-06-22T09:55:32,208][ERROR][o.o.i.i.ManagedIndexCoordinator] [node-1] Failed to clear ManagedIndexMetadata for index uuid: [SzDddFrKSTWIiRc9PHyYOA#metadata], failureMessage: [.opendistro-ism-config] IndexNotFoundException[no such index [.opendistro-ism-config]]
[2021-06-22T09:55:32,208][ERROR][o.o.i.i.ManagedIndexCoordinator] [node-1] Failed to clear ManagedIndexMetadata for index uuid: [1pVfrCqrQP68naDdJS7yrg#metadata], failureMessage: [.opendistro-ism-config] IndexNotFoundException[no such index [.opendistro-ism-config]]
[2021-06-22T09:55:32,208][ERROR][o.o.i.i.ManagedIndexCoordinator] [node-1] Failed to clear ManagedIndexMetadata for index uuid: [76S3Gp03SliRpp4RvH_hrA#metadata], failureMessage: [.opendistro-ism-config] IndexNotFoundException[no such index [.opendistro-ism-config]]

**Other plugins installed**
ALL
**To Reproduce**
[...]

PR #89, opened 2021-06-30
rolling upgrade issues fixes

bowenlan-amzn:

*Issue #, if available:*
#82 

*Description of changes:*
Fix several issues found during rolling upgrade tests:
1. Policy saved in managed index are not parsed out with non-empty `policy_id` field. This previously is only for user to do search on it, we treat it as read-only but not needed for parse out.
2. Update metadata in Runner is checking config index mapping schema version now.
3. Coordinator clean metadata throw a lot of error when config index not exist.

*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
@@ -63,6 +63,7 @@ import org.opensearch.common.xcontent.XContentType
 import org.opensearch.index.Index
 import org.opensearch.index.engine.VersionConflictEngineException
 import org.opensearch.index.seqno.SequenceNumbers
+import org.opensearch.indexmanagement.IndexManagementIndices
 import org.opensearch.indexmanagement.IndexManagementPlugin.Companion.INDEX_MANAGEMENT_INDEX
 import org.opensearch.indexmanagement.indexstatemanagement.action.Action
 import org.opensearch.indexmanagement.indexstatemanagement.model.ManagedIndexConfig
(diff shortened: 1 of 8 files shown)


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/ManagedIndexConfig.kt
```diff
@@ -189,6 +189,7 @@ data class ManagedIndexConfig(
                 policySeqNo = policySeqNo,
                 policyPrimaryTerm = policyPrimaryTerm,
                 policy = policy?.copy(
+                    id = policyID ?: NO_ID,
```
thalurur, line 192: we are checking above that policyID to be not null in line 188 - do we need the else assignment here?

dbbaughe: It's a var so compiler assumes it can be String? still in case it got reassigned after the null check

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -598,6 +605,10 @@ object ManagedIndexRunner :
         lastUpdateResult: UpdateMetadataResult? = null
     ): UpdateMetadataResult {
         var result = UpdateMetadataResult()
+        if (!imIndices.attemptUpdateConfigIndexMapping()) {
+            logger.error("Failed to update config index mapping.")
+            return result
```
dbbaughe, line 610: What is the behavior of the calling code when it gets an empty UpdateMetadataResult?

bowenlan-amzn: not empty, `UpdateMetadataResult` is default to indicate metadata not saved.

dbbaughe (approved):

PR #91, opened 2021-06-30
Choose the rollup settings to use based on cluster configuration

thalurur:
*Issue #, if available:*
https://github.com/opensearch-project/index-management/issues/85

*Description of changes:*
Use the existing cluster configuration checker to see if there is ODFE and OS version of IndexManagement plugin installed and if so use the old settings when creating rollup target index

*CheckList:*
[X ] Commits are signed per the DCO using --signoff


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/SkipExecution.kt
```diff
@@ -77,13 +80,24 @@ class SkipExecution(
...
+                                if (nodePlugin.name == "opendistro-index-management" ||
+                                    nodePlugin.name == "opendistro_index_management"
+                                ) {
+                                    legacyVersionSet.add(nodePlugin.version)
+                                }
                             }
                         }
 
-                    if (versionSet.size > 1) {
+                    if ((versionSet.size + legacyVersionSet.size) > 1) {
```
dbbaughe, line 92: Was this needed? Like if there is a cluster that has new nodes coming up with the new opensearch plugin then shouldn't that be reported as a new nodePlugin version and it would be added to the versionSet?

Or is it possible for us to have IM OpenSearch x.y.z and IM OpenDistro x.y.z where x.y.z are the same on both and you need to rely on the actual plugin name?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/SkipExecution.kt
```diff
@@ -77,13 +80,24 @@ class SkipExecution(
...
                         flag = true
-                        logger.info("There are multiple versions of Index Management plugins in the cluster: $versionSet")
+                        logger.info("There are multiple versions of Index Management plugins in the cluster: [$versionSet, $legacyVersionSet]")
                     } else flag = false
+
+                    if (versionSet.isNotEmpty() && legacyVersionSet.isNotEmpty()) {
```
dbbaughe, line 97: So if versionSet is not empty but legacyVersionSet is empty we say it's not a mixedCluster? That seems wrong. What's the definition of mixedCluster we are going with here? Purely for ODFE vs OS? Or should mixed cluster mean the actual versions (like it normally does right now)

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/SkipExecution.kt
thalurur: I think this flag is tracking if there are multiple versions of the plugins installed, multiple version referring to the number or the name.

If cluster has IM OpenSearch x.y.z and OpenDistro x.y.z then its considered two different versions. And similarly OpenSearch x.y.z and OpenSearch a.b.c is also two different versions

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/SkipExecution.kt
thalurur: may be I can rename this to be more specific, mixedCluster intention is to track if its a combination of ODFE and OS nodes

thalurur: I am not sure what is a good name for it - probably `openSearchUpgrade`?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/SkipExecution.kt
dbbaughe: My question is more can "**IM OpenSearch x.y.z and OpenDistro x.y.z**" even happen? We changed the version with the first release of OpenSearch no?

thalurur: not with this upgrade no, in future may be?

dbbaughe: Hmm sure maybe when we reach version 6.x-7.x OpenSearch which could theoretically match the ODFE 6.x-7.x versions, but I don't believe it's possible for someone with ODFE 6.x-7.x to upgrade directly to those OS versions... they have to upgrade to the OS 1.x most likely.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/SkipExecution.kt
dbbaughe: From this point on this is opensearch so I can you can just call it hasLegacyPlugin or hasODFEPlugin or something and put a comment explaining what it is, represents, and for. Just not mixedCluster since that can imply something else.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/SkipExecution.kt
thalurur: are you suggesting to only have one versionSet object to hold all versionsets? I can update it, but we would still need to check the case where ODFE and OS IM plugin is installed on nodes of the cluster and can use some other boolean to hold that information

dbbaughe: No, I understand you need to check specifically for ODFE plugins for the logic you want to implement. I'm just pointing out that you prob don't need to use it here to set this specific flag to true because simply checking for any difference in versions would be a superset of also checking if OS and ODFE plugins exist in cluster because you should logically never have the same versions for OS and ODFE in the same cluster. You just mainly want it for the further check where you are setting the other flag to true to represent OS and ODFE for the upgrade.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/SkipExecution.kt
thalurur: sure, will do that

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/SkipExecution.kt
dbbaughe: Scratch this -- talked with Ravi and I didn't have the top code part expanded so didn't see it was hardcoded checking for opensearch versions. Thought the versionSet was all index management versions.

bowenlan-amzn (approved):

Issue #75, opened 2021-06-09
Support granular security


praveensameneni:
The user with API permission, with delete rollup, can delete all rollup jobs. We don’t have the way to stop them from deleting rollup jobs created and used by other user.
Use case:
User Bob created a rollup job - Job 123
Bob has access to all the API’s on the job -
POST/PUT/CREATE
GET
MODIFY

User Jane has GET permissions to the API, but should not have access to the job123 created by Bob

should not have access to job123

PR #113, opened 2021-08-05
Storing user information as part of the job when security plugin is installed

thalurur:
*Issue #, if available:*
https://github.com/opensearch-project/index-management/issues/75

*Description of changes:*
Storing user object as part of the job when information is available and setting the user roles in threadcontext while running the ISM policy actions to ensure the user who created managed indices have the necessary permissions before running the action

*CheckList:*
[  ] Commits are signed per the DCO using --signoff

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
@@ -53,6 +53,7 @@ import org.opensearch.cluster.block.ClusterBlockException
 import org.opensearch.cluster.metadata.IndexMetadata
 import org.opensearch.cluster.service.ClusterService
 import org.opensearch.common.component.LifecycleListener
+import org.opensearch.common.regex.Regex
 import org.opensearch.common.settings.Settings
 import org.opensearch.common.unit.TimeValue
 import org.opensearch.index.Index
@@ -61,14 +62,13 @@ import org.opensearch.index.query.QueryBuilders
 import org.opensearch.indexmanagement.IndexManagementIndices
 import org.opensearch.indexmanagement.IndexManagementPlugin
 import org.opensearch.indexmanagement.IndexManagementPlugin.Companion.INDEX_MANAGEMENT_INDEX
-import org.opensearch.indexmanagement.indexstatemanagement.model.ISMTemplate
 import org.opensearch.indexmanagement.indexstatemanagement.model.ManagedIndexConfig
 import org.opensearch.indexmanagement.indexstatemanagement.model.ManagedIndexMetaData
+import org.opensearch.indexmanagement.indexstatemanagement.model.Policy
 import org.opensearch.indexmanagement.indexstatemanagement.model.coordinator.ClusterStateManagedIndexConfig
 import org.opensearch.indexmanagement.indexstatemanagement.model.coordinator.SweptManagedIndexConfig
-import org.opensearch.indexmanagement.indexstatemanagement.opensearchapi.filterNotNullValues
-import org.opensearch.indexmanagement.indexstatemanagement.opensearchapi.getPolicyToTemplateMap
 import org.opensearch.indexmanagement.indexstatemanagement.opensearchapi.mgetManagedIndexMetadata
+import org.opensearch.indexmanagement.indexstatemanagement.opensearchapi.parsePolicies
 import org.opensearch.indexmanagement.indexstatemanagement.settings.ManagedIndexSettings.Companion.AUTO_MANAGE
 import org.opensearch.indexmanagement.indexstatemanagement.settings.ManagedIndexSettings.Companion.COORDINATOR_BACKOFF_COUNT
 import org.opensearch.indexmanagement.indexstatemanagement.settings.ManagedIndexSettings.Companion.COORDINATOR_BACKOFF_MILLIS
@@ -273,6 +273,7 @@ class ManagedIndexCoordinator(
      */
     @OpenForTesting
     suspend fun sweepClusterChangedEvent(event: ClusterChangedEvent) {
+        logger.info("test:      processing cluster change event")
         // indices delete event
         var removeManagedIndexReq = emptyList<DocWriteRequest<*>>()
         var indicesToClean = emptyList<Index>()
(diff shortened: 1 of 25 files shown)


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -283,7 +291,7 @@ object ManagedIndexRunner :
         }
 
         val state = policy.getStateToExecute(managedIndexMetaData)
-        val action: Action? = state?.getActionToExecute(clusterService, scriptService, client, settings, managedIndexMetaData)
+        val action: Action? = state?.getActionToExecute(clusterService, scriptService, client, settings, managedIndexMetaData.copy(user = managedIndexConfig.user))
```
bowenlan-amzn, line 294: What is this for?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/getpolicy/GetPoliciesResponse.kt
```diff
@@ -72,7 +72,7 @@ class GetPoliciesResponse : ActionResponse, ToXContentObject {
                         .field(_ID, policy.id)
                         .field(_SEQ_NO, policy.seqNo)
                         .field(_PRIMARY_TERM, policy.primaryTerm)
-                        .field(Policy.POLICY_TYPE, policy, XCONTENT_WITHOUT_TYPE)
+                        .field(Policy.POLICY_TYPE, policy, XCONTENT_WITHOUT_TYPE_AND_USER)
```
bowenlan-amzn, line 75: Supposed to not showing the user object from this request or temporary thing?

thalurur: Not showing it for now, will have to change this once all the APIs are changed

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
thalurur: So the User information is available inside the `Step` for Rollup ISM Action the User object needs to be populated from the managed Index

src/main/kotlin/org/opensearch/indexmanagement/util/SecurityUtils.kt
```diff
@@ -0,0 +1,39 @@
...
+
+class SecurityUtils {
+    companion object {
+        const val INTERNAL_REQUEST = "index_management_plugin_internal_user"
+        val DEFAULT_INJECT_ROLES: List<String> = listOf("all_access", "AmazonES_all_access")
+        /**
+         * Helper method to build the user object either from the threadContext or from the requested user.
+         */
+        fun buildUser(threadContext: ThreadContext, requestedUser: User? = null): User? {
+            if (threadContext.getTransient<Boolean>(INTERNAL_REQUEST) != null && threadContext.getTransient<Boolean>(INTERNAL_REQUEST)) {
```
dbbaughe, line 26: I think these getTransients can throw exceptions, should we wrap these in a try catch or does the caller do it?

src/main/kotlin/org/opensearch/indexmanagement/opensearchapi/OpenSearchExtensions.kt
```diff
@@ -199,3 +212,42 @@ fun <T> XContentParser.parseWithType(
...
+    override val key: CoroutineContext.Key<*>
+        get() = Key
+    val injector = InjectSecurity(id, settings, threadContext)
+
+    /**
+     * Before the thread executes the coroutine we want the thread context to contain user roles so they are used when executing the code inside
+     * the coroutine
+     */
+    override fun updateThreadContext(context: CoroutineContext) {
+        logger.info("Setting security context in thread ${Thread.currentThread().name} for job $id")
```
dbbaughe, line 235: I think this (updateThreadContext) gets called every time coroutine resumes right? Could be some log spam w/ info log level, especially once thousands of managed indices running and then add in rollups and transforms etc.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/util/RestHandlerUtils.kt
```diff
@@ -106,13 +110,25 @@ data class FailedIndex(val name: String, val uuid: String, val reason: String) :
...
+    user: User? = null
 ): XContentBuilder {
-    return XContentFactory.jsonBuilder()
-        .startObject()
-        .startObject(ManagedIndexConfig.MANAGED_INDEX_TYPE)
-        .optionalTimeField(ManagedIndexConfig.LAST_UPDATED_TIME_FIELD, Instant.now())
-        .field(ManagedIndexConfig.CHANGE_POLICY_FIELD, changePolicy)
-        .endObject()
-        .endObject()
+    if (user == null) {
```
dbbaughe, line 116: Seems like a lot of duplicationg below, can you just set val builder = ... and then if (user == null) builder.field(...etc) and return builder?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/indexpolicy/TransportIndexPolicyAction.kt
```diff
@@ -175,11 +176,13 @@ class TransportIndexPolicyAction @Inject constructor(
         }
 
         private fun putPolicy() {
-            request.policy.copy(schemaVersion = IndexUtils.indexManagementConfigSchemaVersion)
```
dbbaughe, line 178: Hmm seems like you caught a bug here too, what was the impact? Schema version not getting updated?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
```diff
@@ -273,6 +273,7 @@ class ManagedIndexCoordinator(
      */
     @OpenForTesting
     suspend fun sweepClusterChangedEvent(event: ClusterChangedEvent) {
+        logger.info("test:      processing cluster change event")
```
dbbaughe, line 276: Clean up

dbbaughe (commented):
I saw we have user saved on policies, managed index configs, and managed index metadata?
Why do we need the user on all three versus just on the policy?

src/main/kotlin/org/opensearch/indexmanagement/util/SecurityUtils.kt
thalurur: let me check, I haven't seen any errors with getTransient, Alerting and other plugins are using the getTransient as part of every transport action with out any wrapper.

thalurur:
> I saw we have user saved on policies, managed index configs, and managed index metadata?
> Why do we need the user on all three versus just on the policy?

We are note persisting user information in managed index metadata, its just part of the instance object that is passed to Step. We are persisting it in Policy and ManagedIndex as top level fields. Also when policy is copied over to ManagedIndex we are not storing the User object inside the ManagedIndex.Policy path.
The user field in ManagedIndex and Policy will be used to filter the data when fetching this for read APIs

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
thalurur: sure, somehow missed it

src/main/kotlin/org/opensearch/indexmanagement/opensearchapi/OpenSearchExtensions.kt
thalurur: right, I will make this debug log so its not logged in production by default

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/util/RestHandlerUtils.kt
thalurur: sure, I actually have `optionalUserField` which is what I want to use eventually. Currently didn't want to return the user field as part of the REST API output at the moment, for now will do what you suggested to reduce code duplication

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/indexpolicy/TransportIndexPolicyAction.kt
thalurur: yes, it is a bug. Earlier we weren't setting the schema version to be latest version but to 1 by default and if I set a random number in input for schema_version that is persisted

dbbaughe (approved):

PR #115, opened 2021-08-11
Storing user object in all APIs and enabling filter of response based on user

thalurur:

*Issue #, if available:*
https://github.com/opensearch-project/index-management/issues/75

*Description of changes:*
* Storing user information populated by security plugin in the APIs
   * For transforms, rollups, policies there is a high level attribute called `user` which stores the user information
   * For managed indices, the attached policies of the managed index will have the `user` information
   * None of the indexed job metadata has the `user` information 
* Using the stored user information to run the background jobs - rollups, transforms, ism jobs
   * The user information is loaded into thread context to do this and this is only done for specific transport actions called as part of the background job - e.g. when writing data to ism config index the threadcontext will not have user information but when reading data from the source index in rollup job will have the user information in threadcontext. This will ensure an less privileged user will not be able to read/write data from other indices they don't have permissions for by creating rollup/transform jobs. Similarly when executing the policy actions the user information is added to thread context
   * Any data that was written prior to this change will not have user information in them and they are regarded as jobs with full permissions
* Enabling new setting to filter stored information based on the appropriate use permissions
    * There is a new setting which enables filtering of data returned by the IndexManagement plugin APIs, if this setting is enabled it is required that all users updating/creating jobs using IM plugin have backend roles. Other wise the request will fail
    * If the setting is enabled a user will only see data that matches their backend roles. All the existing data that is written without backend roles will not be visible to anyone (This is a limitation)

*Call outs:*
* One can no longer create a managed index job with a non existent policy, add policy API will verify that the policy exists and is stored with the managed index at the time of creation
* Permission check for managed indices is implemented based on a custom permission implemented as part of new Transport action called `ManagedIndexAction`, i.e if user has this permission on an index they will be able to manage the index - this check is implemented in AddPolicy, removePolicy, ChangePolicy, RetryFailedIndex API and the cluster event sweeper to determine the policy that should be applied to an index 
* Explain API will always fail to return if user attempts to fetch indices they don't have permissions for - i.e on every index in the response the user needs to have the permission
* User information stored in the jobs is not being returned as part of the REST API responses 

*CheckList:*
[  ] Commits are signed per the DCO using --signoff


src/main/resources/mappings/opendistro-ism-config.json
```diff
@@ -614,43 +614,6 @@
               }
             }
           }
-        },
-        "user": {
```
dbbaughe, line 618: Was this added in previous PR? Or from a long time ago?

thalurur: last PR few days ago, not included in any release

build.gradle
```diff
@@ -154,6 +154,7 @@ dependencies {
     compile group: 'commons-codec', name: 'commons-codec', version: '1.13'
     compile "org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}"
     compile "org.jetbrains.kotlin:kotlin-stdlib-common:${kotlin_version}"
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:${kotlin_version}"
```
dbbaughe, line 157: nit: Are we adding a whole dependency just for the "use" helper function? Is it not possible to just copy that helper function over? Seems a bit overkill.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/ManagedIndexMetaData.kt
```diff
@@ -64,7 +65,8 @@ data class ManagedIndexMetaData(
     val info: Map<String, Any>?,
     val id: String = NO_ID,
     val seqNo: Long = SequenceNumbers.UNASSIGNED_SEQ_NO,
-    val primaryTerm: Long = SequenceNumbers.UNASSIGNED_PRIMARY_TERM
+    val primaryTerm: Long = SequenceNumbers.UNASSIGNED_PRIMARY_TERM,
+    val user: User? = null
```
dbbaughe, line 69: Can we add a comment here explaining what this is used for and how it is not to be streamed/written anywhere.

thalurur: Will do

src/test/kotlin/org/opensearch/indexmanagement/IndexManagementIndicesIT.kt
```diff
@@ -133,7 +133,8 @@ class IndexManagementIndicesIT : IndexStateManagementRestTestCase() {
 
         val managedIndexConfig = getExistingManagedIndexConfig(index)
         assertNull("Change policy is not null", managedIndexConfig.changePolicy)
-        assertNull("Policy has already initialized", managedIndexConfig.policy)
+        // We are storing the policy immediately as part of the add policy API
+        // assertNull("Policy has already initialized", managedIndexConfig.policy)
```
dbbaughe, line 137: Do we need to keep the commented out code and this comment?

build.gradle
thalurur: Yes, its added only for `use` method. Can copy the method instead - do we follow any rule of thumb on when to copy vs add dependency?

src/test/kotlin/org/opensearch/indexmanagement/IndexManagementIndicesIT.kt
thalurur: this was a miss, will remove

build.gradle
dbbaughe: I don't think there's a rule of thumb we follow.. I'd just look at dependency size (+ maintenance overhead, maintainer activity, etc.) vs what we're using from it. If it's just a simple method that we can import ourselves then might be able to not add a whole dependency, but if you find you're having to pull in a lot of things it's fine to add.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
```diff
@@ -354,31 +358,61 @@ class ManagedIndexCoordinator(
...
+                val searchRequest = SearchRequest().indices(indexName)
+                    .source(SearchSourceBuilder.searchSource().size(1))
+                withClosableContext(IndexManagementSecurityContext("ApplyPolicyOnIndexCreation", settings, threadPool.threadContext, policy.user)) {
+                    val response: SearchResponse = client.suspendUntil { search(searchRequest, it) }
+                }
+            } catch (e: OpenSearchSecurityException) {
+                logger.debug("Skipping applying policy ${policy.id} on $indexName as the policy user is missing perimissions", e)
+                return false
+            } catch (e: Exception) {
+                // Ignore other exceptions
```
dbbaughe, line 411: Should we still add a debug log here?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
```diff
@@ -91,25 +110,82 @@ class TransportAddPolicyAction @Inject constructor(
...
+        private lateinit var policy: Policy
         private val indicesToAdd = mutableMapOf<String, String>() // uuid: name
         private val failedIndices: MutableList<FailedIndex> = mutableListOf()
 
         fun start() {
-            ismIndices.checkAndUpdateIMConfigIndex(object : ActionListener<AcknowledgedResponse> {
-                override fun onResponse(response: AcknowledgedResponse) {
-                    onCreateMappingsResponse(response)
+            if (!validateUserConfiguration(user, filterByEnabled, actionListener)) {
+                return
```
dbbaughe, line 123: This and other places that do same check -> if you just return then I believe you are leaving an open connection waiting for a response? Can you use the actionListener to respond and then return?

dbbaughe: Ah ignore, saw implementation of function, seems internally it does that.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
bowenlan-amzn: Instead of ignore, the best choice to skip this `findMatchingPolicy` of this index, let the next sweep do this again. Otherwise we are assuming the `OpenSearchSecurityException` is always the first exception we would get during a search call.

src/main/kotlin/org/opensearch/indexmanagement/opensearchapi/OpenSearchExtensions.kt
```diff
@@ -266,3 +300,14 @@ class IndexManagementSecurityContext(
         injector.close()
     }
 }
+
+suspend fun <T> withClosableContext(
```
bowenlan-amzn, line 304: I am curious about this. When do you see the leak of injected threadContext which needs this finally injector.close

src/main/kotlin/org/opensearch/indexmanagement/util/SecurityUtils.kt
```diff
@@ -27,13 +34,92 @@ class SecurityUtils {
...
             }
-
             val injectedUser: User? = User.parse(threadContext.getTransient<String>(ConfigConstants.OPENSEARCH_SECURITY_USER_INFO_THREAD_CONTEXT))
             return if (injectedUser == null) {
                 null
             } else {
                 User(injectedUser.name, injectedUser.backendRoles, injectedUser.roles, injectedUser.customAttNames)
             }
         }
+
```
dbbaughe, line 44: Do we have unit tests for all the below new methods?

dbbaughe:
Few top level comments:

* Lots of changes in APIs, just confirm we are not introducing any breaking changes by removing or renaming fields or changing models etc.
* Lots of new behavior added for security, but I know we have issues running security tests locally w/ the local integ Cluster that runs. Can we still add integration tests that get filtered out if remote cluster is not used? i.e. we can still write integration tests, but they only run if isRemoteCluster && isSecurityEnabled or something.

src/main/kotlin/org/opensearch/indexmanagement/util/SecurityUtils.kt
```diff
@@ -27,13 +34,92 @@ class SecurityUtils {
...
+        /**
+         * If filterBy is enabled and security is disabled or if filter by is enabled and backend role are empty
+         * we should prevent users from scheduling new jobs
+         */
+        fun <T> validateUserConfiguration(user: User?, filterEnabled: Boolean, actionListener: ActionListener<T>): Boolean {
+            if (filterEnabled) {
+                if (user == null) {
```
bowenlan-amzn, line 51: Curious: why `user==null` meaning security is disabled? Is the total thread context empty, or just user is null?

thalurur: The user stored in the policy we are checking is null, meaning the policy was created when security plugin is disabled in the cluster

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
thalurur: Can add a debug log

thalurur: @bowenlan-amzn per my understanding security plugin call is the first call made for transport action and it will fail with security exception. Is there any case where this is not true?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/removepolicy/TransportRemovePolicyAction.kt
```diff
@@ -91,24 +107,26 @@ class TransportRemovePolicyAction @Inject constructor(
...
-            client.admin()
-                .cluster()
-                .state(
-                    clusterStateRequest,
-                    object : ActionListener<ClusterStateResponse> {
-                        override fun onResponse(response: ClusterStateResponse) {
-                            val indexMetadatas = response.state.metadata.indices
-                            indexMetadatas.forEach {
-                                indicesToRemove.putIfAbsent(it.value.indexUUID, it.key)
+            client.threadPool().threadContext.stashContext().use {
```
dbbaughe, line 110: Curious, does this throw an exception if user is trying to access an index they can't access?

src/main/kotlin/org/opensearch/indexmanagement/opensearchapi/OpenSearchExtensions.kt
thalurur: Lot of times when testing locally the thread context is not cleared out

src/main/kotlin/org/opensearch/indexmanagement/util/SecurityUtils.kt
bowenlan-amzn: But this user is from the thread context, right?

bowenlan-amzn:
If an existing ISM job doesn't have user info saved, what are the ways for user to adopt the new security model?
By calling change policy API?

And same question for Rollup and Transform jobs without user.

src/main/kotlin/org/opensearch/indexmanagement/rollup/RollupRunner.kt
```diff
@@ -257,7 +259,12 @@ object RollupRunner :
                 }
             }
 
-            when (val result = rollupMapperService.attemptCreateRollupTargetIndex(updatableJob, clusterConfigurationProvider.hasLegacyPlugin)) {
+            val result = withClosableContext(
+                IndexManagementSecurityContext(job.id, settings, threadPool.threadContext, job.user)
```
dbbaughe, line 263: Do we need to create new IndexManagementSecurityContext each time (here and below in while loop etc) or can we just reuse a single one?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
bowenlan-amzn: I am not sure, but if that's the case, that sounds good.

src/main/kotlin/org/opensearch/indexmanagement/util/SecurityUtils.kt
thalurur: oh sorry, I misread the comment place. This user is from the thread context. Yes, ``user==null`` means the security plugin is disabled

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/explain/TransportExplainAction.kt
```diff
@@ -68,10 +78,21 @@ class TransportExplainAction @Inject constructor(
...
     transportService: TransportService,
     actionFilters: ActionFilters,
+    val clusterService: ClusterService,
+    val settings: Settings,
     val xContentRegistry: NamedXContentRegistry
 ) : HandledTransportAction<ExplainRequest, ExplainResponse>(
     ExplainAction.NAME, transportService, actionFilters, ::ExplainRequest
 ) {
+
+    @Volatile private var filterByEnabled = IndexManagementSettings.FILTER_BY_BACKEND_ROLES.get(settings)
```
bowenlan-amzn, line 88: Not used for now?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/removepolicy/TransportRemovePolicyAction.kt
thalurur: No, this is a miss. Updating the check for ChangePolicy and RemovePolicy

src/main/kotlin/org/opensearch/indexmanagement/util/SecurityUtils.kt
thalurur: no, don't have the tests, will add

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/explain/TransportExplainAction.kt
thalurur: Since managed indices are indices not thinking about using the filter setting like I am for other APIs, will remove

thalurur:
> If an existing ISM job doesn't have user info saved, what are the ways for user to adopt the new security model?
> By calling change policy API?
> 
> And same question for Rollup and Transform jobs without user.

Yes, updating would be the option

dbbaughe (dismissed):

dbbaughe (approved):

PR #133, opened 2021-08-25
Added functionality for Shrink Action

ansmathu:
*Issue #, if available:*

*Description of changes:*

*CheckList:*
[ ] Commits are signed per the DCO using --signoff

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/WaitForMoveShardsStep.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/WaitForMoveShardsStep.kt
@@ -16,7 +16,6 @@ import org.opensearch.action.admin.indices.stats.IndicesStatsRequest
 import org.opensearch.action.admin.indices.stats.IndicesStatsResponse
 import org.opensearch.action.admin.indices.stats.ShardStats
 import org.opensearch.client.Client
-import org.opensearch.cluster.routing.ShardRouting
 import org.opensearch.cluster.service.ClusterService
 import org.opensearch.index.shard.ShardId
 import org.opensearch.indexmanagement.indexstatemanagement.model.ManagedIndexMetaData
@@ -71,7 +70,8 @@ class WaitForMoveShardsStep(
                 }
                 // TODO: Test if we can make this appear / if we can, fail the action.
                 shardToCheckpointSetMap.entries.forEach {
-                    (_, checkpointSet) -> if (checkpointSet.size > 1) {
+                    (_, checkpointSet) ->
+                    if (checkpointSet.size > 1) {
                         logger.warn("There are shards with varying local checkpoints")
                     }
                 }


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/managedindexmetadata/ShrinkActionProperties.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.opensearch.common.xcontent.ToXContent
+import org.opensearch.common.xcontent.ToXContentFragment
+import org.opensearch.common.xcontent.XContentBuilder
+import org.opensearch.common.xcontent.XContentParser
+import org.opensearch.common.xcontent.XContentParserUtils
+
+data class ShrinkActionProperties(
+    val nodeName: String? = null,
+    val targetIndexName: String? = null,
+    val targetNumShards: Int? = null,
```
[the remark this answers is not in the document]
thalurur: action properties don't have all the fields always, but that is not true here some of the shrink action properties as some of these are always present. Change the fields that are always present in ShrinkActionProperties to be non-nullable.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
```diff
@@ -12,45 +12,313 @@
...
+    val context: JobExecutionContext
 ) : Step(name, managedIndexMetaData) {
     private val logger = LogManager.getLogger(javaClass)
     private var stepStatus = StepStatus.STARTING
     private var info: Map<String, Any>? = null
-
+    private var numTargetShards: Int? = null
+    private var numOriginalShards: Int? = null
+    private var nodeName: String? = null
+    private var shrinkTargetIndexName: String? = null
```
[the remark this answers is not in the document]
thalurur: Do you think you can define a ``var ShrinkActionProperties`` and set that in `getUpdatedManagedIndexMetadata` instead of defining of each individual ShrinkActionProperties as nullable var?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
```diff
@@ -12,45 +12,313 @@
...
+                suitableNodes.add(sizeNodeTuple)
+            }
+        }
+        return suitableNodes
+    }
+
+    @SuppressWarnings("ReturnCount")
+    private fun getNumTargetShards(numOriginalShards: Int, indexSize: Long): Int {
+        // case where user specifies a certain number of shards in the target index
+        if (config.numNewShards != null) return getGreatestFactorLessThan(numOriginalShards, config.numNewShards)
```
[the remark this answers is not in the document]
thalurur: This is not obvious, probably need to be called out in the documentation that this input is an upper bound and not used exactly.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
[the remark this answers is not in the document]
ansmathu: Done

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/managedindexmetadata/ShrinkActionProperties.kt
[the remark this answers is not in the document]
ansmathu: All the node related fields are left as nullable (in the case that the node doesn't acquire). This is includes nodeName and the 3 lock fields. The targetIndexName and the targetIndexShards are updated to be non-nullable.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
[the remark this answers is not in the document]
ansmathu: For sure

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
```diff
@@ -12,45 +12,313 @@
...
+            if (lock == null) {
+                info = mapOf("message" to getNoAvailableNodesMessage(managedIndexMetaData.index))
+                stepStatus = StepStatus.CONDITION_NOT_MET
+                return this
+            }
+            // set index to read only
+            setToReadOnly()
+            // move the shards
+            nodeName = lock!!.resource["node_name"] as String
+            moveIndexToNode(nodeName!!)
```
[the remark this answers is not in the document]
ansmathu: Done

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
```diff
@@ -12,45 +12,336 @@
...
+                return this
+            }
+            // set index to read only
+            setToReadOnly()
+            // move the shards
+            val nodeName = lock.resource["node_name"] as String
+            shrinkActionProperties = ShrinkActionProperties(
+                nodeName,
+                shrinkActionProperties!!.targetIndexName,
+                shrinkActionProperties!!.targetNumShards,
```
thalurur, line 143: can you just use the ``shrinkTargetIndexName`` and ``numTargetShards`` instead?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
```diff
@@ -12,45 +12,336 @@
...
+            // get the number of shards that the target index will have
+            val numTargetShards = getNumTargetShards(numOriginalShards, indexSize)
+            shrinkActionProperties = ShrinkActionProperties(
+                null,
+                shrinkTargetIndexName,
+                numTargetShards,
+                null,
+                null,
+                null
+            )
```
thalurur, line 119: is there any benefit in setting this? if not can we make all the properties for shrinkActionProperties required and set only when we have all since all of those are needed to proceed beyond this step?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
```diff
@@ -12,45 +12,336 @@
...
+
+    private suspend fun setToReadOnly() {
+        val updateSettings = Settings.builder().put(IndexMetadata.SETTING_BLOCKS_WRITE, true).build()
+        issueUpdateAndUnlockIfFail(updateSettings, getReadOnlyFailedMessage(managedIndexMetaData.index))
+    }
+
+    private suspend fun moveIndexToNode(node: String) {
+        val allocationSettings = Settings.builder().put(ROUTING_SETTING, node).build()
+        issueUpdateAndUnlockIfFail(allocationSettings, getAllocationFailedMessage(managedIndexMetaData.index, node))
+    }
```
thalurur, line 167: can you merge these two calls to one?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
```diff
@@ -12,45 +12,336 @@
...
+        numOriginalShards: Int
+    ): PriorityQueue<Tuple<Long, String>> {
+        val nodesStatsReq = NodesStatsRequest().addMetric(OS_METRIC)
+        val nodeStatsResponse: NodesStatsResponse = client.admin().cluster().suspendUntil { nodesStats(nodesStatsReq, it) }
+        val nodesList = nodeStatsResponse.nodes
+        val comparator = kotlin.Comparator { o1: Tuple<Long, String>, o2: Tuple<Long, String> -> o1.v1().compareTo(o2.v1()) }
+        val nodesWithSpace = PriorityQueue<Tuple<Long, String>>(comparator)
+        for (node in nodesList) {
+            val memLeftInNode = node.os!!.mem.free.bytes
+            val totalNodeMem = node.os!!.mem.total.bytes
```
thalurur, line 222: instead of using `!!` can you check if not null and execute the iteration?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
```diff
@@ -12,45 +12,336 @@
...
+            val bufferSize = ByteSizeValue(buffer * totalNodeMem)
+            val requiredBytes = (2 * indexSize) + bufferSize.bytes
+            if (memLeftInNode > requiredBytes) {
+                val memLeftAfterTransfer: Long = memLeftInNode - requiredBytes
+                nodesWithSpace.add(Tuple(memLeftAfterTransfer, node.node.name))
```
thalurur, line 227: why do we need to store the `memLeftAfterTransfer` I don't see it being used again anywhere outside? Can the return type changed to just be the node names?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
```diff
@@ -12,45 +12,336 @@
...
+                null
+            )
+
+            // get the nodes with enough memory
+            val suitableNodes = findSuitableNodes(statsResponse, indexSize, bufferPercentage, numOriginalShards)
+            var nodesString = ""
+            for (node in suitableNodes) {
+                nodesString += "\t${node.v2()}\n"
+            }
+            logger.info("The nodes that ${managedIndexMetaData.index} can shrink onto are as follows:\n$nodesString")
```
thalurur, line 127: is this needed to be logged as info?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/managedindexmetadata/ShrinkActionProperties.kt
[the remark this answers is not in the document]
thalurur: it makes sense to make all the fields required right? we are not going to proceed to other steps unless we have all the information and this information is only stored so next steps can execute?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/util/StepUtils.kt
```diff
@@ -0,0 +1,94 @@
...
+        if (!released) {
+            logger.warn("Lock not released on failure")
+        }
+    } catch (e: RemoteTransportException) {
+        throw e
+    }
+}
+
+fun getShrinkLockModel(
+    managedIndexMetaData: ManagedIndexMetaData,
```
thalurur, line 43: can you just pass the shrinkActionProperties?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
ansmathu: Sorry -- this is something that I had in the demo branch that must have made it into this branch during the merge. Will remove.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/util/StepUtils.kt
```diff
@@ -0,0 +1,94 @@
...
+        lockCreationInstant,
+        WaitForMoveShardsStep.MOVE_SHARDS_TIMEOUT_IN_SECONDS,
+        false,
+        shrinkActionProperties.lockSeqNo!!,
+        shrinkActionProperties.lockPrimaryTerm!!
+    )
+}
+
+@SuppressWarnings("LongParameterList")
+fun getShrinkLockModel(
```
thalurur, line 64: I can't search for this in the PR? is this needed - also why two methods doing the same thing line 42 and line 64

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
ansmathu: The memLeftAfterTransfer is used to order the PriorityQueue. The nodes with the most memory after the transfer are picked first. Then in acquireLockOnNode iterates in that order to attempt to get the locks on those nodes.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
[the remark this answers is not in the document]
thalurur: this seems to be missed?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
thalurur: I see, I get that we are using this to order the data in the priority queue initially so its fine to keep it here.

But once its ordered we don't need to return both values right in line 230 or am I misunderstanding something here? - We can just return the list of nodes in that order from this method and acquireLockOnNode can use this list in that order to get the lock?

thalurur: I see, I get that we are using this to order the data in the priority queue initially so its fine to keep it here.

But once its ordered we don't need to return both values right in line 230 or am I misunderstanding something here? - We can just return the list of nodes in that order from this method and acquireLockOnNode can use this list in that order to get the lock?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
```diff
@@ -12,45 +12,336 @@
...
+        const val RESOURCE_TYPE = "shrink"
+        fun getSuccessMessage(index: String, node: String) = "Successfully started moving the shards of $index to $node."
+        fun getReadOnlyFailedMessage(index: String) = "Shrink failed because $index could not be set to read only."
+        fun getAllocationFailedMessage(index: String, node: String) = "Shrink failed because the shards of $index could not be moved $node."
+        fun getNoAvailableNodesMessage(index: String) =
+            "There are no available nodes for $index to move to to execute a shrink. Delaying until node becomes available."
+        fun getFailureMessage(index: String) = "Shrink failed to start moving shards of $index."
+        fun getIndexExistsMessage(index: String, newIndex: String) = "Shrink failed on $index because $newIndex already exists."
+        fun getUnsafeFailure(index: String) = "Shrink failed because $index has no replicas and force_unsafe is not set to true."
+        fun getOnePrimaryShardFailure(index: String) = "Shrink failed because $index only has one primary shard."
```
thalurur, line 345: remove $index from all these, its redundant as the information is displayed with index in explain API

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptShrinkStep.kt
```diff
@@ -48,7 +98,8 @@ class AttemptShrinkStep(
 
     companion object {
         const val name = "attempt_shrink_step"
-        const val FIVE_MINUTES_IN_MILLIS = 1000 * 60 * 5 // how long to wait for the force merge request before moving on
-        const val FIVE_SECONDS_IN_MILLIS = 1000L * 5L // delay
+        fun getSuccessMessage(index: String, newIndex: String) = "Shrink started on $index. $newIndex currently being populated."
+        fun getIndexHealthNotGreenMessage(index: String) = "Shrink delayed because $index health is not green."
+        fun getFailureMessage(index: String) = "Shrink failed on $index when sending shrink request."
```
thalurur, line 103: remove $index information as its redundant

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/readonly/SetReadOnlyStep.kt
```diff
@@ -57,12 +56,8 @@ class SetReadOnlyStep(
...
     override suspend fun execute(): SetReadOnlyStep {
         try {
-            val updateSettingsRequest = UpdateSettingsRequest()
-                .indices(indexName)
-                .settings(Settings.builder().put(SETTING_BLOCKS_WRITE, true))
-            val response: AcknowledgedResponse = client.admin().indices()
-                .suspendUntil { updateSettings(updateSettingsRequest, it) }
-
+            val updateSettings = Settings.builder().put(SETTING_BLOCKS_WRITE, true).build()
+            val response: AcknowledgedResponse = issueUpdateSettingsRequest(client, managedIndexMetaData, updateSettings)
```
thalurur, line 60: This seems to be incorrect? the index name is not passed to the request?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
thalurur: Also we need to set the index name in the request right?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
```diff
@@ -12,45 +12,336 @@
...
+    }
+
+    private suspend fun moveIndexToNode(node: String) {
+        val allocationSettings = Settings.builder().put(ROUTING_SETTING, node).build()
+        issueUpdateAndUnlockIfFail(allocationSettings, getAllocationFailedMessage(managedIndexMetaData.index, node))
+    }
+
+    private suspend fun issueUpdateAndUnlockIfFail(settings: Settings, failureMessage: String) {
+        try {
+            val response: AcknowledgedResponse = issueUpdateSettingsRequest(client, managedIndexMetaData, settings)
```
thalurur, line 171: we need to set the indexName in the request right?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptShrinkStep.kt
```diff
@@ -12,28 +12,78 @@
...
+            // check status of cluster health
+            if (response.isTimedOut) {
+                stepStatus = StepStatus.CONDITION_NOT_MET
+                info = mapOf("message" to getIndexHealthNotGreenMessage(managedIndexMetaData.index))
+                return this
+            }
+            val targetIndexName = managedIndexMetaData.actionMetaData.actionProperties.shrinkActionProperties.targetIndexName
+            val aliases = config.aliases
+            val req = ResizeRequest(targetIndexName, managedIndexMetaData.index)
+            req.targetIndexRequest.settings(
```
thalurur, line 65: we should also put the allocation setting for target index to be selected shrink node right?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
ansmathu: Yeah that is possible. I could add a for loop at the end that just gives the first values.

thalurur: we are already doing that in line 253 it looks like just add the node name of the tuple?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/readonly/SetReadOnlyStep.kt
ansmathu: The index name is set in the issueUpdateIndexRequest.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/WaitForMoveShardsStep.kt
```diff
@@ -12,28 +12,88 @@
...
+            }
+            val numShardsLeft = numPrimaryShards - numShardsOnNode
+            checkTimeOut(numShardsLeft, nodeToMoveOnto)
+            return this
+        } catch (e: RemoteTransportException) {
+            releaseShrinkLock(managedIndexMetaData, context, logger)
+            info = mapOf("message" to getFailureMessage(indexName))
+            stepStatus = StepStatus.FAILED
+            return this
+        }
```
thalurur, line 96: for all the steps please add catch block for generic Exception and set the fail message

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/WaitForMoveShardsStep.kt
```diff
@@ -48,9 +108,35 @@ class WaitForMoveShardsStep(
...
     }
 
+    private suspend fun checkTimeOut(numShardsLeft: Int, nodeToMoveOnto: String) {
+        val timeFromActionStarted: Duration = Duration.between(getActionStartTime(managedIndexMetaData), Instant.now())
+        val timeOutInSeconds = config.configTimeout?.timeout?.seconds ?: MOVE_SHARDS_TIMEOUT_IN_SECONDS
+        // Get ActionTimeout if given, otherwise use default timeout of 12 hours
+        stepStatus = if (timeFromActionStarted.toSeconds() > timeOutInSeconds) {
+            logger.error(
+                "The shards of "
+            )
```
thalurur, line 118: seems to be incomplete?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
ansmathu: The index name is set in the issueUpdatedSettingsRequest.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
```diff
@@ -12,45 +12,338 @@
...
-
+    private var shrinkActionProperties: ShrinkActionProperties? = null
     override fun isIdempotent() = true
 
-    @Suppress("TooGenericExceptionCaught", "ComplexMethod")
+    @Suppress("TooGenericExceptionCaught", "ComplexMethod", "ReturnCount", "LongMethod")
     override suspend fun execute(): AttemptMoveShardsStep {
-        return this
+        try {
+            checkTimeOut()
```
thalurur, line 74: curious: why is this needed?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
```diff
@@ -12,45 +12,338 @@
...
+                info = mapOf("message" to getOnePrimaryShardFailure())
+                stepStatus = StepStatus.FAILED
+                return this
+            }
+            // Get the size of the index
+            val statsRequest = IndicesStatsRequest().indices(indexName)
+            val statsResponse: IndicesStatsResponse = client.admin().indices().suspendUntil {
+                stats(statsRequest, it)
+            }
+            val indexSize = statsResponse.total.getStore()!!.sizeInBytes
```
thalurur, line 115: can you do a null check instead of `!!`?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
```diff
@@ -12,45 +12,338 @@
...
+            val lock = acquireLockOnNode(suitableNodes)
+            if (lock == null) {
+                logger.info("${managedIndexMetaData.index} could not find available node to shrink onto.")
+                info = mapOf("message" to getNoAvailableNodesMessage())
+                stepStatus = StepStatus.CONDITION_NOT_MET
+                return this
+            }
+            // set index to read only
+            // move the shards
+            val nodeName = lock.resource["node_name"] as String
```
thalurur, line 131: can you reuse the constant defined in other step?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
```diff
@@ -12,45 +12,338 @@
...
+            val suitableNodes = findSuitableNodes(statsResponse, indexSize, bufferPercentage, numOriginalShards)
+            // iterate through the nodes and try to acquire a lock on those nodes
+            val lock = acquireLockOnNode(suitableNodes)
+            if (lock == null) {
+                logger.info("${managedIndexMetaData.index} could not find available node to shrink onto.")
+                info = mapOf("message" to getNoAvailableNodesMessage())
+                stepStatus = StepStatus.CONDITION_NOT_MET
+                return this
+            }
+            // set index to read only
```
thalurur, line 129: minor: cleanup?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
```diff
@@ -12,45 +12,338 @@
...
+    private suspend fun issueUpdateAndUnlockIfFail(settings: Settings, failureMessage: String) {
+        try {
+            val response: AcknowledgedResponse = issueUpdateSettingsRequest(client, managedIndexMetaData, settings)
+            if (!response.isAcknowledged) {
+                stepStatus = StepStatus.FAILED
+                info = mapOf("message" to failureMessage)
+            }
+        } catch (e: Exception) {
+            handleException(e, failureMessage)
+            val lock = getShrinkLockModel(
```
thalurur, line 168: can you check shrinkActionProperties is non null before calling the release lock to prevent the `!!`?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
ansmathu: I didn't want the action to run forever if an available node is not found.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptShrinkStep.kt
```diff
@@ -12,28 +12,92 @@
...
-    @Suppress("TooGenericExceptionCaught", "ComplexMethod")
+    @Suppress("TooGenericExceptionCaught", "ComplexMethod", "ReturnCount")
     override suspend fun execute(): AttemptShrinkStep {
-        return this
+        try {
+            if ((managedIndexMetaData.actionMetaData?.actionProperties?.shrinkActionProperties == null)) {
+                info = mapOf("message" to "Metadata not properly populated")
+                stepStatus = StepStatus.FAILED
+                return this
+            }
```
thalurur, line 53: optional: do you wanna do this outside of try block so the you don't need to if `shrinkActionProperties` is not null in catch blocks

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/WaitForMoveShardsStep.kt
```diff
@@ -12,28 +12,92 @@
...
+            return this
+        } catch (e: RemoteTransportException) {
+            if (managedIndexMetaData.actionMetaData?.actionProperties?.shrinkActionProperties != null) {
+                releaseShrinkLock(managedIndexMetaData.actionMetaData.actionProperties.shrinkActionProperties, context, logger)
+            }
+            info = mapOf("message" to getFailureMessage())
+            stepStatus = StepStatus.FAILED
+            return this
+        } catch (e: Exception) {
+            info = mapOf("message" to getFailureMessage(), "cause" to "{${e.message}}")
```
thalurur, line 97: missing releasing of lock

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/WaitForShrinkStep.kt
```diff
@@ -12,28 +12,123 @@
...
+            stepStatus = StepStatus.FAILED
+            return this
+        } catch (e: Exception) {
+            if (managedIndexMetaData.actionMetaData?.actionProperties?.shrinkActionProperties != null) {
+                releaseShrinkLock(
+                    managedIndexMetaData.actionMetaData.actionProperties.shrinkActionProperties,
+                    context,
+                    logger
+                )
+                info = mapOf("message" to getGenericFailureMessage(), "cause" to "{${e.message}}")
```
thalurur, line 103: minor: you can move this out of if block and remove the else?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/AttemptMoveShardsStep.kt
ansmathu: I didn't want this step to go on forever if no nodes are available.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/managedindexmetadata/ShrinkActionProperties.kt
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+
+    enum class ShrinkProperties(val key: String) {
+        SHRINK_NODE_NAME("shrink_node_name"),
+        SHRINK_TARGET_INDEX_NAME("shrink_target_index_name"),
+        SHRINK_TARGET_NUM_SHARDS("shrink_target_num_shards"),
+        SHRINK_LOCK_SEQ_NO("shrink_lock_seq_no"),
+        SHRINK_LOCK_PRIMARY_TERM("shrink_lock_primary_term"),
+        SHRINK_LOCK_EPOCH_SECOND("shrink_lock_epoch_second")
```
thalurur, line 94: optional: since these are encapsulated in shrink_properties I think we can drop the shrink prefix to these

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/WaitForMoveShardsStep.kt
```diff
@@ -12,28 +12,91 @@
...
+            stepStatus = StepStatus.FAILED
+            return this
+        }
+        try {
+            val indexStatsRequests: IndicesStatsRequest = IndicesStatsRequest().indices(managedIndexMetaData.index)
+            val response: IndicesStatsResponse = client.admin().indices().suspendUntil { stats(indexStatsRequests, it) }
+            val numPrimaryShards = clusterService.state().metadata.indices[managedIndexMetaData.index].numberOfShards
+            val nodeToMoveOnto = managedIndexMetaData.actionMetaData.actionProperties.shrinkActionProperties.nodeName
+            var numShardsOnNode = 0
+            val shardToCheckpointSetMap: HashMap<ShardId, HashSet<Long>> = HashMap()
```
thalurur, line 61: minor: can you use kotlin types? `val shardToCheckpointSetMap: MutableMap<String, MutableSet<Long>> = mutableMapOf()`

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/WaitForMoveShardsStep.kt
```diff
@@ -12,28 +12,91 @@
...
+            for (shard: ShardStats in response.shards) {
+                val checkpoint = shard.seqNoStats!!.localCheckpoint
```
thalurur, line 63: can we do explicit null check instead and fail if null?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/WaitForMoveShardsStep.kt
```diff
@@ -12,28 +12,91 @@
...
+                val routingInfo: ShardRouting = shard.shardRouting
+                val shardId = shard.shardRouting.shardId()
+                if (shardId in shardToCheckpointSetMap.keys && !shardToCheckpointSetMap[shardId]!!.equals(checkpoint)) {
+                    shardToCheckpointSetMap[shardId]!!.add(checkpoint)
+                    val checkPointSet = shardToCheckpointSetMap[shardId]
+                    logger.warn("There are shards with varying local checkpoints for $shardId. The checkpoints are $checkPointSet.")
+                } else {
+                    shardToCheckpointSetMap[shardId] = HashSet()
+                    shardToCheckpointSetMap[shardId]!!.add(checkpoint)
+                }
```
thalurur, line 73: Is this correct - the value in map is a set and you are comparing to just a long?

`!shardToCheckpointSetMap[shardId]!!.equals(checkpoint)`

To prevent usage of `!!` this can be refactored a little

```
val checkpoints = shardToCheckpointSetMap.getOrDefault(key, mutableSetOf())
checkpoints.add(checkpoint")
shardToCheckpointSetMap.put(key, checkpoints)
```

and after the for loop you can check if there are any shardIds that have multiple checkpoints to see if there are varying checkpoints

```
shardToCheckpointSetMap.entries.forEach { (shardId, checkPointSet) ->
            if (checkPointSet.count() > 1) {
logger.warn("There are shards with varying local checkpoints for $shardId. The checkpoints are $checkPointSet.")}
        }
```

thalurur (dismissed):

thalurur (approved):

arnitolog:
Hi, 
will this action be available in the upcoming release? I see that it was merged into the feature branch a long time ago.

PR #117, opened 2021-08-13
ISM/Notification channel support

dbbaughe:
*Issue #, if available:*

*Description of changes:*

*CheckList:*
[ ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/index-management/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #136, opened 2021-08-31
Removes support for notification plugin

dbbaughe:
Goes through previous PR that added support for notification plugin and removes it
https://github.com/opensearch-project/index-management/pull/117

It's not a complete git revert as there are still some things we want to/can keep.

*CheckList:*
[x] Commits are signed per the DCO using --signoff


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -372,14 +369,8 @@ object ManagedIndexRunner :
...
-                    withClosableContext(
-                        IndexManagementSecurityContext(
-                            managedIndexConfig.id, settings, threadPool.threadContext, managedIndexConfig.policy.user
-                        )
-                    ) {
-                        // if the policy has no error_notification this will do nothing otherwise it will try to send the configured error message
-                        publishErrorNotification(policy, executedManagedIndexMetaData)
-                    }
+                    // if the policy has no error_notification this will do nothing otherwise it will try to send the configured error message
+                    publishErrorNotification(policy, executedManagedIndexMetaData)
```
bowenlan-amzn, line 373: Should we wrap this using the role injection wrapper?

dbbaughe: Since we are reverting back to previous notification implementation it is now just a construction of the http client and directly publishing. We needed the role injection previously because it was using transport actions and users needed the permissions for those transport actions to use the new channels in notification plugin. It will be added back once we revert this back to supporting the new notification plugin.

bowenlan-amzn (approved):

PR #141, opened 2021-09-09
Uses published daily snapshot dependencies

dbbaughe:

*CheckList:*

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,9 +41,9 @@ buildscript {
     }
 
     repositories {
-        mavenLocal()
         mavenCentral()
         maven { url "https://plugins.gradle.org/m2/" }
+        maven { url "https://aws.oss.sonatype.org/content/repositories/snapshots" }
     }
 
     dependencies {
@@ -182,7 +182,7 @@ dependencies {
 }
 
 repositories {
-    mavenLocal()
+    maven { url "https://aws.oss.sonatype.org/content/repositories/snapshots" }
 }
 
 plugins.withId('java') {
@@ -192,7 +192,6 @@ plugins.withId('java') {
 plugins.withId('org.jetbrains.kotlin.jvm') {
     compileKotlin.kotlinOptions.jvmTarget = compileTestKotlin.kotlinOptions.jvmTarget = "1.8"
     compileKotlin.dependsOn ktlint
-    
 }
 
 javadoc.enabled = false // turn off javadoc as it barfs on Kotlin code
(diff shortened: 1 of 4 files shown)


dbbaughe:
That is interesting.... link checker fails because of link at bottom of rfc
https://github.com/opensearch-project/index-management/blob/main/docs/rfc.md

But why didn't it fail for previous PRs...

bowenlan-amzn (approved):

build.gradle
```diff
@@ -182,7 +182,7 @@ dependencies {
 }
 
 repositories {
-    mavenLocal()
```
saratvemulapalli, line 185: As part of the CI/CD we are trying to run integration/bwc tests which qualifies the bundle and allows the maven artifacts to be published. 
It would rely on `mavenlocal()` to run these tests.

dbbaughe: Ah I see, I'll add it back w/ a comment explaining what it's for as it didn't seem needed in our repo.

dbbaughe: Did you need mavenLocal in the buildscript repositories too?

saratvemulapalli: Yeah still trying to figure out if there is any other way. But seems like `mavenlocal()` could be a simpler solution at this point. 
The CI pipelines are in development we are learning through the process, so we didn't know either. 

Could you point me to the buildscript ?

dbbaughe: https://github.com/opensearch-project/index-management/blob/main/build.gradle#L43-L47

I removed it from here too.

When you do your CI builds are you building release or snapshot? Because if you're also trying to build SNAPSHOT and I add this mavenLocal back.. then I have to add it to the top of the repositories so that it makes sure to check there first to find it (otherwise it'll find it first in our SNAPSHOT repository). But, if we do make mavenLocal first... now all devs/contributors need to make sure they clean out their local maven repositories of OpenSearch related artifacts if they want to make sure they are actually getting it from the SNAPSHOT repo which is a bit annoying.

saratvemulapalli: Yeah buildscript fetches the build-tools i.e `opensearchplugin` and it would depend on `mavenlocal()` too. 
That is a great point. I think it would be SNAPSHOT repository but i'll get back on this.

dblock: I really think you should back this out of 1.1 and make the change in 1.2. Branch for 1.1. We didn't give our maven publishing any bake time and the bundle build doesn't need more issues right now.

dbbaughe: @dblock This commit was not included in our 1.1 branch, and is not in our 1.x branch yet either. It's only in main.

build.gradle
```diff
@@ -41,9 +41,9 @@ buildscript {
     }
 
     repositories {
-        mavenLocal()
```
dblock, line 44: We need to preserve `mavenLocal` here for bundle build.

PR #147, opened 2021-09-15
Adds implementation for the delay feature in rollup jobs

downsrob:
### Description
Adds implementation for the delay parameter in rollup jobs. Previously, the delay parameter was included in the api and stored with the job, but implementation was never added to use this delay when searching for windows of data to be used in the rollup job. In this PR, the rollups exclude the data from the delay time period, and the delay is passed to the job scheduler to handle delaying the actual execution.

As these changes rely on the delay parameter in the job scheduler plugin, which has not yet been merged, the tests are expected to fail until that is merged. All tests pass on my local machine using the updated job scheduler instance from [this PR.](https://github.com/opensearch-project/job-scheduler/pull/61) Included in the changed files is the snapshot zip of the job scheduler from that same PR.

*CheckList:*


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -485,7 +485,7 @@ object ManagedIndexRunner :
     private suspend fun updateJobInterval(managedIndexConfig: ManagedIndexConfig, jobInterval: Int) {
         try {
             val updatedManagedIndexConfig = managedIndexConfig
-                .copy(jobSchedule = IntervalSchedule(getIntervalStartTime(managedIndexConfig), jobInterval, ChronoUnit.MINUTES))
+                .copy(jobSchedule = IntervalSchedule(getIntervalStartTime(managedIndexConfig), jobInterval, ChronoUnit.MINUTES, 0))
```
thalurur, line 488: since there are two constructors do we need to specify the delay here? can we not just not set the delay at all

src/main/kotlin/org/opensearch/indexmanagement/rollup/RollupSearchService.kt
```diff
@@ -26,6 +26,7 @@
 
 package org.opensearch.indexmanagement.rollup
 
+import kotlinx.coroutines.delay
```
thalurur, line 29: remove? seems to be unused

src/test/kotlin/org/opensearch/indexmanagement/TestHelpers.kt
```diff
@@ -61,9 +61,9 @@ fun randomEpochMillis(): Long = OpenSearchRestTestCase.randomLongBetween(0, Inst
 
 fun randomInstant(): Instant = Instant.ofEpochMilli(randomEpochMillis())
 
-fun randomCronSchedule(): CronSchedule = CronSchedule(randomCronExpression(), OpenSearchRestTestCase.randomZone())
+fun randomCronSchedule(): CronSchedule = CronSchedule(randomCronExpression(), OpenSearchRestTestCase.randomZone(), 0)
```
thalurur, line 64: same as above - can we not just use the constructor without delay here? any reason to set it to 0?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
downsrob: To maintain backwards compatibility, if the schedule objects use the constructor without the delay then the delay value is set to null and it is excluded from the xcontent, which is good for the plugins which don't expect the delay parameter in the xcontent. Unfortunately it also means that we always need to use the constructor which includes the delay if we want the delay to be included in the xcontent.
The only way I can think to get around this confusing strategy would be to make other objects which extend the IntervalSchedule and JobSchedule objects but which also include delay, do you think that would be better?

src/main/kotlin/org/opensearch/indexmanagement/rollup/util/RollupUtils.kt
```diff
@@ -94,14 +95,16 @@ fun isRollupIndex(index: String, clusterState: ClusterState): Boolean {
...
     val query = if (metadata.continuous != null) {
         RangeQueryBuilder(this.getDateHistogram().sourceField)
             .from(metadata.continuous.nextWindowStartTime.toEpochMilli(), true)
             .to(metadata.continuous.nextWindowEndTime.toEpochMilli(), false)
             .format(DATE_FIELD_EPOCH_MILLIS_FORMAT)
     } else {
-        MatchAllQueryBuilder()
+        RangeQueryBuilder(this.getDateHistogram().sourceField)
+            .to(max(0L, Instant.now().toEpochMilli() - (job.delay ?: 0)), false)
+            .format(DATE_FIELD_EPOCH_MILLIS_FORMAT)
```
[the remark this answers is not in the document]
downsrob: I believe that we need this for the non-continuous delayed rollups. 
If you start a non-continuous rollup at 10 AM with a delay of 1 hour, then at 11 AM it should roll up everything from epoch to 10 AM, not epoch to 11 AM, right?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
thalurur: If delay is not scheduled in the rollup and if we use the non-delay param constructor then the xcontent of rollup will not delay field in the interval/cron schedule. But if delay is set in the rollup then the interval/cron schedule is updated to have the delay field and the xcontent of rollup will contain the delay field. 

Is the issue in first scenario the xcontent won't have delay? I am not sure why that is an issue since there is no delay specified in that case, is there anything I am missing here on what/why the issue?

PR #148, opened 2021-09-16
Removing default integtest.sh.

dblock:
Coming from https://github.com/opensearch-project/opensearch-build/issues/497, removing default integtest.sh.


bowenlan-amzn (approved):

bowenlan-amzn:
Thanks for taking care of this.

PR #153, opened 2021-09-23
Adds cluster setting to configure index state management jitter

downsrob:
Signed-off-by: Clay Downs downsrob@amazon.com

*Description of changes:* Adds cluster setting to allow users to enable and configure the jitter setting for index state management jobs. The purpose of the jitter is to add a randomized delay to the start time of ISM jobs to make sure that all indices do not become active at the same time. The jitter is stored as a double and represents the maximum percentage of the job interval to add to the job start time. The default jitter is 0.6, and the cluster setting resides under plugins.index_state_management.jitter.

The job scheduler has another configurable cluster setting, jitter_limit, which represents the maximum jitter which may be applied to the job. The jitter cluster setting may be higher than this limit, but jitter times will be capped to the jitter_limit percentage of the job interval. The job scheduler jitter_limit default is 0.6. 

At this time, modifying the jitter cluster setting will only apply the changed jitter to newly created jobs. 

*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
@@ -75,6 +75,7 @@ import org.opensearch.indexmanagement.indexstatemanagement.settings.ManagedIndex
 import org.opensearch.indexmanagement.indexstatemanagement.settings.ManagedIndexSettings.Companion.COORDINATOR_BACKOFF_COUNT
 import org.opensearch.indexmanagement.indexstatemanagement.settings.ManagedIndexSettings.Companion.COORDINATOR_BACKOFF_MILLIS
 import org.opensearch.indexmanagement.indexstatemanagement.settings.ManagedIndexSettings.Companion.INDEX_STATE_MANAGEMENT_ENABLED
+import org.opensearch.indexmanagement.indexstatemanagement.settings.ManagedIndexSettings.Companion.JITTER
 import org.opensearch.indexmanagement.indexstatemanagement.settings.ManagedIndexSettings.Companion.JOB_INTERVAL
 import org.opensearch.indexmanagement.indexstatemanagement.settings.ManagedIndexSettings.Companion.METADATA_SERVICE_ENABLED
 import org.opensearch.indexmanagement.indexstatemanagement.settings.ManagedIndexSettings.Companion.SWEEP_PERIOD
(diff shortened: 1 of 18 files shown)


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/settings/ManagedIndexSettings.kt
```diff
@@ -179,5 +180,14 @@ class ManagedIndexSettings {
             Setting.Property.NodeScope,
             Setting.Property.Dynamic
         )
+
+        val JITTER: Setting<Double> = Setting.doubleSetting(
+            "plugins.index_state_management.jitter",
+            DEFAULT_JITTER,
+            0.0,
+            Double.MAX_VALUE,
```
thalurur, line 188: the max should be 1.0 right? Don't think it make sense to set a jitter greater than original schedule

downsrob: Good point. I was thinking that the jitter_limit parameter of the job scheduler would constrain the jitter value anyways so I didn't need to add an upper limit, but adding the limitation here would be better as it would throw an error

src/main/resources/mappings/opendistro-ism-config.json
```diff
@@ -614,6 +614,9 @@
               }
             }
           }
+        },
+        "jitter": {
+          "type": "double"
```
thalurur, line 619: do you think its better to not store the jitter? rather always get it from cluster setting instead that way the jitter is dynamic and changing the setting will update the jitter for all the jobs

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
```diff
@@ -328,7 +333,8 @@ class ManagedIndexCoordinator(
                                 indexUuid,
                                 policy.id,
                                 jobInterval,
-                                policy
+                                policy,
+                                jobJitter
```
thalurur, line 337: this is just one way of creating the managed index jobs, they can be created using the attach policy API too

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/ManagedIndexConfig.kt
```diff
@@ -79,6 +80,10 @@ data class ManagedIndexConfig(
 
     override fun getLockDurationSeconds(): Long = 3600L // 1 hour
 
+    override fun getJitter(): Double? {
+        return jobJitter
+    }
```
thalurur, line 85: do we want to read the cluster setting value here and return it instead?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
thalurur: see the PR actually has the API covered too, resolving

PR #157, opened 2021-09-30
Updates index management version to 1.2

downsrob:
Signed-off-by: Clay Downs downsrob@amazon.com
*Issue #, if available:*
https://github.com/opensearch-project/opensearch-build/issues/663
*Description of changes:*
Updates index management version to 1.2. Changes the gradle build file, as well as the github test and build workflows.
*CheckList:*


dblock:
Since this is to include IM into the 1.2.0 bundle, can we please address https://github.com/opensearch-project/index-management/issues/155 part of this as well? See https://github.com/opensearch-project/opensearch-plugins/blob/main/BUILDING.md#consume-maven-artifacts-in-order that was just added to explain the order of maven dependencies required.

dbbaughe:
@dblock Yes, we plan to merge in the common utils and job scheduler 1.2 version changes first. Let those get published to the snapshot repository, and then do the IM changes. We can add back the mavenLocal repo at that time too.

dblock:
Ok cool, I want to gate adding into the 1.2 manifest any plugins that haven't moved to using maven. This way we don't have to chase them separately.

dbbaughe:
Sure, Clay is making the changes in common utils and job scheduler to depend on the staging repo

dbbaughe:
@downsrob Ah you're in a bit of a predicament now... looks like it's trying to find the alerting notification subproject 1.2 on maven.. so you *could* go and now bump alerting main to 1.2 and whatever else is needed to get that notification subproject published to the staging repo.. but at the same time.. I *think* we're trying to remove it in 1.2, but we haven't merged back in the support for notification plugin yet in Index Management as we are waiting to confirm it'll go out in 1.2.

dbbaughe (approved):

PR #165, opened 2021-10-04
Adds setting to search all rollup jobs on a target index

downsrob:

*Description of changes:*
A user can set up rollup jobs such a way that the job can read data from set of indices and write to a target index.
Multiple jobs can be writing to the same target index. The users can then retrieve the data from target index by doing a _search  however, when executing search we pick data from only one rollup job on the target index instead of all data. The reason to do this was to prevent duplicate data being returned since we have no constraints on what type of jobs are writing to the target index.

Some users have valid use cases where they want to search data from all the rollup jobs in a target index. To support those use cases, this PR offers a rollup cluster setting which controls this behavior - selecting one or all jobs for rollup search. 

*CheckList:*

### DIFF
--- a/src/test/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptorIT.kt
+++ b/src/test/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptorIT.kt
@@ -663,4 +663,152 @@ class RollupInterceptorIT : RollupRestTestCase() {
             rollupAggRes.getValue("min_passenger_count")["value"]
         )
     }
+
+    fun `test rollup search all jobs`() {
+        generateNYCTaxiData("source_rollup_search_all_jobs_1")
+        generateNYCTaxiData("source_rollup_search_all_jobs_2")
+        val targetIndex = "target_rollup_search_all_jobs"
+        val rollupHourly = Rollup(
+            id = "hourly_basic_term_query_rollup_search_all",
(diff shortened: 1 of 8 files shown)


src/test/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptorIT.kt
```diff
@@ -663,4 +663,152 @@ class RollupInterceptorIT : RollupRestTestCase() {
...
+        val trueAggSum = rawAgg1Res.getValue("sum_passenger_count")["value"] as Double + rawAgg2Res.getValue("sum_passenger_count")["value"] as Double
+        assertNotEquals(
+            "Source and rollup index did not return different sum results",
+            trueAggSum, rollupAggResSingle.getValue("sum_passenger_count")["value"]
+        )
+        val trueAggCount = rawAgg1Res.getValue("value_count_passenger_count")["value"] as Int + rawAgg2Res.getValue("value_count_passenger_count")["value"] as Int
+        assertNotEquals(
+            "Source and rollup index did not return different value_count results",
+            trueAggCount, rollupAggResSingle.getValue("value_count_passenger_count")["value"]
+        )
```
thalurur, line 792: May be I am misreading this, only one rollup job should be picked here since default setting is false right? then why are we doing assert of sum of both rollup jobs equal to search response

Issue #179, opened 2021-11-02
Allow out of band rollovers


dbbaughe:
**Is your feature request related to a problem? Please describe.**
Allow users to rollover an index without breaking ISM managed job for that index when it also tries to roll it over.

PR #180, opened 2021-11-02
Allows out of band rollovers on an index without causing ISM to fail

dbbaughe:

*Issue #, if available:*
https://github.com/opensearch-project/index-management/issues/179

*Description of changes:*
Previously when a user would rollover an index that ISM is managing and waiting to rollover, ISM would enter a bad state and cause ISM to do multiple rollovers of aliases.
There was a fix put in place to block this from happening where we'd also check to confirm the job performing the rollover for a specific index also confirms that index is still the write alias and fail otherwise.
This is good to prevent the first issue, but now users have to deal with failed indices and use the skip rollover setting.
This allows out of band rollovers to happen without affecting ISM, ISM will check to see if the specific index has already been rolled over for the given alias, and if so treat it as successful.

*CheckList:*


thalurur (approved):

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -72,18 +72,16 @@ class AttemptRolloverStep(
...
         }
 
-        // If we have already rolled over this index then fail as we only allow an index to be rolled over once
-        if (managedIndexMetaData.rolledOver == true) {
-            logger.warn("$indexName was already rolled over, cannot execute rollover step")
-            stepStatus = StepStatus.FAILED
-            info = mapOf("message" to getFailedDuplicateRolloverMessage(indexName))
-            return this
-        }
-
```
bowenlan-amzn, line 82: If this is removed, then this rolledover attribute is not useful anymore and can be cleaned up, right?

dbbaughe: If this is the only place it was actively used.. then technically yes, but at the same time users will have metadata that has this boolean on it regardless so we can't remove it completely from our codebase if our parsing is strict and throws an exception when it sees an unrecognized field.

Issue #125, opened 2021-08-23
Add backwards compatibility tests for automation


VachaShah:
Add backwards compatibility tests for the plugin to automate end-to-end upgrade paths for faster releases. Developer documentation for implementing bwc tests and hooking them to CI: https://github.com/opensearch-project/opensearch-plugins/blob/main/TESTING.md#backwards-compatibility-testing

dbbaughe:
Marked as closed since Erics PR was merged.

PR #181, opened 2021-11-02
Added ISM policy backwards compatibility test

lobdelle:
*Issue #, if available:* https://github.com/opensearch-project/index-management/issues/125 

*Description of changes:*
- Added ISM Policy backwards compatibility tests for mixed cluster, rolling upgrade and full restart scenarios
- Updates DEVELOPER_GUIDE to add instructions for running bwc tests
- Adds bwc tests to the multi-node GitHub Actions tests workflow

*CheckList:*


build.gradle
```diff
@@ -306,6 +309,12 @@ integTest {
         jvmArgs '-agentlib:jdwp=transport=dt_socket,server=n,suspend=y,address=8000'
     }
 
+    if (System.getProperty("tests.rest.bwcsuite") == null) {
+        filter {
+            excludeTestsMatching "org.opensearch.alerting.bwc.*IT"
```
dbbaughe, line 314: There ain't no alerting in here 😛

lobdelle: Whoops, fixed.

build.gradle
```diff
@@ -319,6 +328,170 @@ integTest {
...
+    dependsOn bundlePlugin
+    doLast {
+        plugins = [
+                provider(new Callable<RegularFile>(){
+                    @Override
+                    RegularFile call() throws Exception {
+                        return new RegularFile() {
+                            @Override
+                            File getAsFile() {
+                                return fileTree("src/test/resources/job-scheduler").getSingleFile()
```
qreshi, line 385: Do we know if we can resolve the current Job Scheduler plugin that the current project version will be running with by fetching it from maven or somewhere else instead of supplying it under `test/resources`? My concern is that you'll need to update that artifact for the newest version as `main` progress to keep the "current" case of the BWC tests accurate which might be bothersome.

@dbbaughe Any thoughts?

.github/workflows/multi-node-test-workflow.yml
```diff
@@ -31,3 +31,7 @@ jobs:
         with:
           name: logs
           path: build/testclusters/integTest-*/logs/*
+      - name: Run IM Backwards Compatibility Tests
+        run: |
+          echo "Running backwards compatibility tests..."
+          ./gradlew bwcTestSuite -Dtests.security.manager=false
```
qreshi, line 37: This was mentioned in the Alerting BWC PR as well but it looks like the tests succeed even without including `-Dtests.security.manager=false` so we can probably exclude it to be consistent with our other gradle test task runs

src/test/kotlin/org/opensearch/indexmanagement/bwc/IndexManagementBackwardsCompatibilityIT.kt
```diff
@@ -0,0 +1,185 @@
...
+                }
+            }
+            ClusterType.UPGRADED -> "_nodes/plugins"
+        }
+    }
+
+    @Throws(Exception::class)
+    private fun createBasicPolicy() {
+        val indexName = "test_bwc_index"
+        val legacyPolicyString = """
```
qreshi, line 110: This had to be done in the Alerting BWC tests because the Monitor had changed enough in the new versions that the `toXContent` response was not forwards compatible if being used in a request against the old version (ODFE 1.13.1).

If this is not the case for ISM, I'd recommend calling `randomPolicy()` and then doing [policy.toHttpEntity()](https://github.com/opensearch-project/index-management/blob/ff055dcc9e90e9c40618134040a32e59936e8da1/src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/IndexStateManagementRestTestCase.kt#L426) to provide it in the request below to keep the variability in the tests.

lobdelle: Gave it a try, looks like this is also the case for policies. Specifically it didn't like the `user` parameter in the Policy.

qreshi: If you do still want to use `randomPolicy()`, it looks like the `toXContent()` for Policy does have a param to exclude user:

https://github.com/opensearch-project/index-management/blob/c0816691ba827669ab4b6f2f9a4857a4fd20e97b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/Policy.kt#L93

So you could use this:
https://github.com/opensearch-project/index-management/blob/c0816691ba827669ab4b6f2f9a4857a4fd20e97b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/util/RestHandlerUtils.kt#L45

And create a util method to get an HttpEntity without user for Policy (assuming something like that doesn't exist already), something like:
```kotlin
Policy.toHttpEntityWithoutUser(): HttpEntity {
    val builder = XContentFactory.jsonBuilder()
    val jsonString = this.toXContent(builder, XCONTENT_WITHOUT_USER).string()
    return StringEntity(toJsonString(), APPLICATION_JSON)
}
```

build.gradle
qreshi: We don't have to address the above in this PR but if we do want to do something about it, let's create a GitHub issue to track it.

qreshi (approved):

.github/workflows/multi-node-test-workflow.yml
```diff
@@ -31,3 +31,7 @@ jobs:
         with:
           name: logs
           path: build/testclusters/integTest-*/logs/*
+      - name: Run IM Backwards Compatibility Tests
```
dbbaughe, line 34: Could we separate this into its own workflow?

src/test/kotlin/org/opensearch/indexmanagement/bwc/IndexManagementBackwardsCompatibilityIT.kt
```diff
@@ -0,0 +1,143 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
```
dbbaughe, line 10: Isn't there a different 2 liner license to use now?

src/test/kotlin/org/opensearch/indexmanagement/ODFERestTestCase.kt
```diff
@@ -117,8 +117,12 @@ abstract class ODFERestTestCase : OpenSearchRestTestCase() {
         }
     }
 
+    open fun preserveODFEIndicesAfterTest(): Boolean = false
+
     @Throws(IOException::class)
     open fun wipeAllODFEIndices() {
+        if (preserveODFEIndicesAfterTest()) return
+
```
dbbaughe, line 125: What was this change for?

.github/workflows/multi-node-test-workflow.yml
qreshi: Just be aware that `./gradlew build` runs all tasks in the task graph so even though we've written to exclude the BWC tests from `integTest`, `./gradlew build` will still pick it up.

src/test/kotlin/org/opensearch/indexmanagement/bwc/IndexManagementBackwardsCompatibilityIT.kt
```diff
@@ -0,0 +1,143 @@
...
+            // increase the timeout here to 90 seconds to handle long waits for a green
+            // cluster health. the waits for green need to be longer than a minute to
+            // account for delayed shards
+            .put(CLIENT_SOCKET_TIMEOUT, "90s")
+            .build()
+    }
+
+    @Throws(Exception::class)
+    @Suppress("UNCHECKED_CAST")
+    fun `test policy backwards compatibility`() {
```
dbbaughe, line 56: I might just not know how this bwc test is working, perhaps there's some under the hood stuff going on from core, but could you explain the intent and purpose of this test and what the test should be doing when executing?

src/test/kotlin/org/opensearch/indexmanagement/ODFERestTestCase.kt
qreshi: Without offering a setting to toggle this to off, it was deleting all the config indices in between runs. We want to preserve indices during the cluster going through upgrades otherwise the Monitor/Policy, etc. made at the beginning won't be around when we come back to verify it.

build.gradle
dbbaughe: Both ways are going to have issues... it's also possible that you point to a released job-scheduler zip somewhere that is pulled down to be installed (from cloudfront or wherever) but it doesn't contain the latest changes in job scheduler you need from the upcoming release (i.e. you make changes in JS and ISM). We can try to think of an alternative approach later on..

src/test/kotlin/org/opensearch/indexmanagement/bwc/IndexManagementBackwardsCompatibilityIT.kt
lobdelle: Added more to this test to cover more. Test is run multiple times on the same cluster between different upgrade stages. Test currently checks that a created ISM policy remains present, and added to a given index. Also implicitly testing use of the relevant backwards compatible APIs. For the BWC tests to pass, this test has to pass before the upgrade, during a mixed state where some nodes are upgraded and others aren't, and after all nodes are upgraded.

qreshi (approved):

dbbaughe (approved):

PR #185, opened 2021-11-04
In explain API not showing the total count to all users

thalurur:
*Issue #, if available:*
N/A

*Description of changes:*
IndexManagement managed indices explain API at the moment show all matching managed indices count to all users, instead just showing the count that users can actually allowed to see.

e.g there are 3 managed indices "a", "b", "c" and user jane is allowed to see only "a"
the output of current explain API looks like 
```
{
   "a": {},
   "total_managed_indices": 3
}
```

after the fix it will look like
```
{
   "a": {},
   "total_managed_indices": 1
}
```

*CheckList:*


dbbaughe:
Should we add a test to catch a regression on this? Or will that be in your security test PR?

thalurur:
> Should we add a test to catch a regression on this? Or will that be in your security test PR?

I am adding the security tests in a different PR this will be checked there

dbbaughe (approved):

PR #203, opened 2021-11-23
Cleanup tests and add ability to choose what to preserve after each test run

thalurur:

*Issue #, if available:*
N/A

*Description of changes:*
* Adding ability to runner to determine what to preserve between integration tests
* Cleanup unused security enabled test code

*CheckList:*

### DIFF
--- a/src/test/kotlin/org/opensearch/indexmanagement/ODFERestTestCase.kt
+++ b/src/test/kotlin/org/opensearch/indexmanagement/ODFERestTestCase.kt
@@ -1,27 +1,6 @@
 /*
+ * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
- *
- * The OpenSearch Contributors require contributions made to
- * this file be licensed under the Apache-2.0 license or a
- * compatible open source license.
- *
- * Modifications Copyright OpenSearch Contributors. See
- * GitHub history for details.
- */
-
-/*
- *   Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
- *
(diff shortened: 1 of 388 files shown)


src/test/kotlin/org/opensearch/indexmanagement/ODFERestTestCase.kt
```diff
@@ -120,20 +162,6 @@ abstract class ODFERestTestCase : OpenSearchRestTestCase() {
...
-        return Settings
-            .builder()
-            .put("http.port", 9200)
-            .put(OPENSEARCH_SECURITY_SSL_HTTP_ENABLED, isHttps())
-            .put(OPENSEARCH_SECURITY_SSL_HTTP_PEMCERT_FILEPATH, "sample.pem")
-            .put(OPENSEARCH_SECURITY_SSL_HTTP_KEYSTORE_FILEPATH, "test-kirk.jks")
-            .put(OPENSEARCH_SECURITY_SSL_HTTP_KEYSTORE_PASSWORD, "changeit")
-            .put(OPENSEARCH_SECURITY_SSL_HTTP_KEYSTORE_KEYPASSWORD, "changeit")
-            .build()
-    }
```
dbbaughe, line 136: Are you sure this isn't needed for when infra runs the tests against their cluster for releases?

thalurur: I am not 100% sure, I can set this only if the security is enabled. I will follow up with the infra team to see if this is needed or not. I only see this being set in four plugins so may be not a requirement but not sure

Issue #152, opened 2021-09-23
Continuous mode in transforms


thalurur:
Even though transforms support a schedule, one can use this schedule to run the job only once. 
Need to support running the job in a continuous mode based on a schedule where new data added to the source indices are transformed in newer runs of transform job

doomma:
@thalurur , would you have any inputs now on a ETA/plan for this enhancement?

JeffBolle:
This should be clarified in the documentation until this issue is resolved.  The UI and documentation lead users (like me) to believe that our transforms should be running continuously.

edwarmatoro:
Hi guys, do you know which version of OpenSearch is this bug fixed? I'm currently using v 1.1.0 and it has the bug; then I moved to 1.1.3 and still, the transform jobs are disabled once they run from the first time.

edwarmatoro:
In case someone else face the same issue, it is resolved. You just need to add the parameter "continuous" : true.

PR #206, opened 2021-12-01
Continuous transforms

downsrob:
*Issue #, if available:*
https://github.com/opensearch-project/index-management/issues/152
*Description of changes:*
This commit adds support for running transform jobs in a continuous mode based on a schedule where new data added to the source indices are transformed in newer runs of transform job. The transform mode may be specified through the "continuous" boolean flag in the request body when creating the transform job. 

At each continuous transform run, the current global checkpoint number, which is the sequence number for which all active shards histories are aligned, is captured for each shard in the source indices. Then, the range of sequence numbers between the previous global checkpoint number and the current global checkpoint number is used to identify the new and modified documents. Only the buckets containing these documents are transformed in each continuous transform iteration. For users with few overlapping buckets between runs, as for users grouping on a document timestamp, time and compute resources are saved by only transforming and indexing the new and modified buckets.

*API changes:*
Creating a transform
```
PUT _plugins/_transform/<transform_id>
Payload additions:
{
    ...
    "continuous": <Optional boolean, determines if the transform job should run every time it is scheduled or just once>
}
```
Describing a transform
```
GET _plugins/_transform/<transform_id>
Payload additions:
{
    ...
    "continuous_stats": { <null if non continuous> 
        "last_timestamp": <timestamp when new data was locked in for the next incremental run>,
        "documents_behind": {
            index_name_1: <number of new documents on this index which have not yet been transformed>
            index_name_2: ...
            ...
        }
    }
}
```

*CheckList:*


dbbaughe:
If some data gets transformed multiple times somehow, let's say the transform runs more than once on a range of data, does that cause any issues or is the transform idempotent and can be run multiple times without affecting the output (besides wasted resource usage potentially)?

downsrob:
@dbbaughe As the modified buckets are re-transformed in full each run, the continuous transform should be idempotent. That scenario would only result in wasted resources.

dbbaughe:
Can switching between continous and non-continous transforms cause any issues to the transform job?
i.e. user creates a non-continuous transform and then updates it to continuous (or the other way around)

src/main/kotlin/org/opensearch/indexmanagement/transform/action/explain/TransportExplainTransformAction.kt
```diff
@@ -123,6 +139,32 @@ class TransportExplainTransformAction @Inject constructor(
                                         else -> actionListener.onFailure(e)
                                     }
                                 }
+
+                                fun getContinuousStats(transform: Transform, metadata: TransformMetadata): ContinuousTransformStats? {
+                                    val shardIDsToGlobalCheckpoint = runBlocking {
```
dbbaughe, line 144: runBlocking blocks the thread, we should never use it in our source code (it's fine for tests).
As we haven't converted our transport classes to use coroutines yet, you might need to have a method signature for passing an action listener.

src/main/kotlin/org/opensearch/indexmanagement/transform/action/explain/TransportExplainTransformAction.kt
```diff
@@ -123,6 +139,32 @@ class TransportExplainTransformAction @Inject constructor(
...
+                                        withClosableContext(
+                                            IndexManagementSecurityContext(
+                                                transform.id,
+                                                settings,
+                                                client.threadPool().threadContext,
+                                                transform.user
+                                            )
+                                        ) {
+                                            transformSearchService.getShardsGlobalCheckpoint(transform.sourceIndex)
+                                        }
```
dbbaughe, line 154: Do we want to make this call as the user stored on transform? That seems like a way for UserA to execute this getShardsGlobalCheckpoint with potentially higher privileges if the user stored is UserB who has more access. Shouldn't we run this with the current users permissions (the one making the REST call)?

downsrob:
> Can switching between continuous and non-continuous transforms cause any issues to the transform job?
> i.e. user creates a non-continuous transform and then updates it to continuous (or the other way around)

Good catch, continuousStats are currently only created when the transform metadata is created for the first time. Non-continuous to continuous would result in the explain rest call not providing the continuous stats. I will add a test case and fix this.
Changing a transform from non continuous to continuous would cause the whole transform to rerun, as the non continuous transform doesn't track sequence numbers, but then should work fine.

dbbaughe:
> Good catch, continuousStats are currently only created when the transform metadata is created for the first time. Non-continuous to continuous would result in the explain rest call not providing the continuous stats. I will add a test case and fix this. Changing a transform from non continuous to continuous would cause the whole transform to rerun, as the non continuous transform doesn't track sequence numbers, but then should work fine.

Do you want this behavior? Otherwise you could just update the IndexTransform transport action and reject any request that tries to change the continuous flag (we do it for rollup).

downsrob:
> Do you want this behavior? Otherwise you could just update the IndexTransform transport action and reject any request that tries to change the continuous flag (we do it for rollup).

I don't see any use case for changing a transform from continuous to non-continuous, and I could go either way on non-continuous to continuous, so I would opt for handling this like rollups, just to maintain consistency. It would be an easy change later on anyways

src/main/kotlin/org/opensearch/indexmanagement/transform/action/explain/TransportExplainTransformAction.kt
thalurur: I think security wise, the validation should be done at the transform level - i.e if the calling user doesn't have permission for transform then it shouldn't even fetch the transform from the query built above - to fetch these stats I think we should follow the pattern of having no user in the threadcontext like we are doing for all other places cause by the time this logic is executed the permission check is already completed

src/main/kotlin/org/opensearch/indexmanagement/common/model/dimension/DateHistogram.kt
```diff
@@ -83,6 +85,18 @@ data class DateHistogram(
...
+    override fun toBucketQuery(bucketKey: Any): AbstractQueryBuilder<*> {
+        if (bucketKey !is Long) {
+            throw IllegalArgumentException("Received invalid date histogram bucket key type [${bucketKey::class}] when Long is expected.")
+        }
+        val interval: Long = DateHistogramInterval(fixedInterval ?: calendarInterval).estimateMillis()
+        return RangeQueryBuilder(sourceField)
+            .from(bucketKey, true)
+            .to(bucketKey + interval, false)
+            .timeZone(timezone.toString())
+            .format("epoch_millis")
```
dbbaughe, line 97: Is the bucketKey coming from user data? If so will it always be "epoch_millis" or will some users have epoch_seconds or nano etc. and this causes a mismatch in the value and format?

downsrob: bucketKey is the key of the bucket returned from the composite aggregation search in TransformSearchService.getModifiedBuckets. That said, I am not sure if the format of the bucket key can be changed by the user source, I will test that

downsrob: DateHistogram keys are internally stored as longs representing epoch milliseconds and the bucket keys will always be in this format

src/main/kotlin/org/opensearch/indexmanagement/transform/TransformSearchService.kt
```diff
@@ -64,7 +78,63 @@ class TransformSearchService(
         }
     }
 
-    suspend fun executeCompositeSearch(transform: Transform, afterKey: Map<String, Any>? = null): TransformSearchResult {
+    suspend fun getShardsGlobalCheckpoint(index: String): Map<ShardId, Long> {
+        val errorMessage = "Failed to get the shards in the source indices"
```
dbbaughe, line 82: nit: Could make these error messages (here and below) static in a companion object so you don't need to assign this every call.

src/main/kotlin/org/opensearch/indexmanagement/transform/TransformSearchService.kt
```diff
@@ -64,7 +78,63 @@ class TransformSearchService(
...
+        try {
+            val request = IndicesStatsRequest().indices(index).clear()
+            val response: IndicesStatsResponse = client.suspendUntil { execute(IndicesStatsAction.INSTANCE, request, it) }
```
dbbaughe, line 85: This will definitely fail sometimes during the course of the continuous transform lifecycle, and it seems like it'll throw the exception up in the runner and fail the job. Does the job end up in a failed state and stop running or do we keep checking? If it ends up stopped, then it's very likely users will get frustrated that the transform jobs aren't resilient to transient failures. Can we introduce some retry/backoff policy here at least?

src/main/kotlin/org/opensearch/indexmanagement/transform/TransformRunner.kt
```diff
@@ -140,46 +157,120 @@ object TransformRunner :
...
+                transformSearchService.getShardsGlobalCheckpoint(transform.sourceIndex)
+            }
+            currentBucketsToTransform = bucketsToTransform.copy(
+                metadata = bucketsToTransform.metadata.copy(
+                    shardIDToGlobalCheckpoint = shardIdToGlobalCheckpoint,
+                    continuousStats = bucketsToTransform.metadata.continuousStats?.copy(lastTimestamp = Instant.now())
+                ),
+                shardsToSearch = getShardsToSearch(oldShardIDToMaxSeqNo, shardIdToGlobalCheckpoint).iterator()
+            )
+            if (currentBucketsToTransform.shardsToSearch!!.hasNext()) currentShard = currentBucketsToTransform.shardsToSearch!!.next()
```
dbbaughe, line 185: Rewrite code to not use !! (here and below in other files, seems to be a lot of uses), this is how accidental NPEs get introduced. !! should only be used in tests.

src/main/kotlin/org/opensearch/indexmanagement/transform/model/TransformMetadata.kt
```diff
@@ -68,7 +73,10 @@ data class TransformMetadata(
         builder.field(STATUS_FIELD, status.type)
         builder.field(FAILURE_REASON, failureReason)
         builder.field(STATS_FIELD, stats)
-
+        if (shardIDToGlobalCheckpoint != null) {
+            builder.field(SHARD_ID_TO_GLOBAL_CHECKPOINT_FIELD, shardIDToGlobalCheckpoint.mapKeys { it.key.toString() })
+        }
```
thalurur, line 78: Do we want to keep some flag to not return this in the API response? not sure if this would be very helpful for users to see when they do explain on the transform job

src/main/kotlin/org/opensearch/indexmanagement/transform/TransformSearchService.kt
```diff
@@ -80,32 +157,68 @@ class TransformSearchService(
...
+
+        fun getSearchServiceRequest(
+            transform: Transform,
+            afterKey: Map<String, Any>? = null,
+            pageSize: Int,
+            modifiedBuckets: MutableSet<Map<String, Any>>? = null
+        ): SearchRequest {
             val sources = mutableListOf<CompositeValuesSourceBuilder<*>>()
-            transform.groups.forEach { group -> sources.add(group.toSourceBuilder()) }
+            transform.groups.forEach { group -> sources.add(group.toSourceBuilder().missingBucket(!transform.continuous)) }
```
thalurur, line 191: curious: what is the use for missingBucket?

src/main/kotlin/org/opensearch/indexmanagement/transform/action/explain/TransportExplainTransformAction.kt
```diff
@@ -129,6 +144,73 @@ class TransportExplainTransformAction @Inject constructor(
                                         else -> actionListener.onFailure(e)
                                     }
                                 }
+
+                                // Recursively adds continuous stats for the next transform in the iterator
+                                private fun addContinuousStats(transformsToSearch: Iterator<Transform>) {
+                                    if (!transformsToSearch.hasNext()) {
+                                        actionListener.onResponse(ExplainTransformResponse(idsToExplain.toMap()))
+                                    } else {
+                                        val currTransform = transformsToSearch.next()
```
thalurur, line 153: Do you think we can have a suspend method in Transform to get the continuous stats? and just call the method in co routine here to help with readability of this logic. So here it would iterate over continuous transforms and for each transform get the continuous stats and add to response?

src/main/kotlin/org/opensearch/indexmanagement/transform/action/explain/TransportExplainTransformAction.kt
```diff
@@ -129,6 +144,73 @@ class TransportExplainTransformAction @Inject constructor(
...
+                                                            )
+                                                        )
+                                                        idsToExplain[currTransform.id] = currentExplain.copy(metadata = newMetadata)
+                                                    }
+                                                    addContinuousStats(transformsToSearch)
+                                                }
+
+                                                override fun onFailure(e: Exception) {
+                                                    log.error("Failed to parse indices stats response", e)
+                                                    actionListener.onFailure(e)
```
thalurur, line 185: hmm.. so if we fail to get one of the continuous transform stats we fail all the entire response? Not sure if that is better or just showing the failure message for just that transform is better

src/main/kotlin/org/opensearch/indexmanagement/transform/TransformRunner.kt
```diff
@@ -140,46 +157,123 @@ object TransformRunner :
...
         }
-        if (validationResult.isValid) {
-            val transformSearchResult = withClosableContext(
-                IndexManagementSecurityContext(transform.id, settings, threadPool.threadContext, transform.user)
-            ) {
-                transformSearchService.executeCompositeSearch(transform, metadata.afterKey)
+        return if (!validationResult.isValid) {
+            val failureMessage = "Failed validation - ${validationResult.issues}"
+            val failureMetadata = transformMetadata.copy(status = TransformMetadata.Status.FAILED, failureReason = failureMessage)
+            transformMetadataService.writeMetadata(failureMetadata, true)
```
thalurur, line 238: why write here? the finally block will write the metadata always right?

src/main/kotlin/org/opensearch/indexmanagement/transform/TransformRunner.kt
```diff
@@ -140,46 +157,123 @@ object TransformRunner :
...
-        ) {
+    private suspend fun getBucketsToTransformIteration(transform: Transform, bucketsToTransform: BucketsToTransform): BucketsToTransform {
+        var currentBucketsToTransform = bucketsToTransform
+        var currentShard = bucketsToTransform.currentShard
+        // If we have not populated the list of shards to search, do so now
+        if (bucketsToTransform.shardsToSearch == null) {
+            val oldShardIDToMaxSeqNo = bucketsToTransform.metadata.shardIDToGlobalCheckpoint
+            val shardIdToGlobalCheckpoint = withTransformSecurityContext(transform) {
+                transformSearchService.getShardsGlobalCheckpoint(transform.sourceIndex)
+            }
```
thalurur, line 177: we don't need to run this as an user, this is required for our internal execution and user doesn't need to have this permission to execute the transform - what do you think?

src/main/kotlin/org/opensearch/indexmanagement/transform/TransformRunner.kt
```diff
@@ -140,46 +157,123 @@ object TransformRunner :
...
+            val shardsToSearch = getShardsToSearch(oldShardIDToMaxSeqNo, shardIdToGlobalCheckpoint).iterator()
+            currentBucketsToTransform = bucketsToTransform.copy(
+                metadata = bucketsToTransform.metadata.copy(
+                    shardIDToGlobalCheckpoint = shardIdToGlobalCheckpoint,
+                    continuousStats = bucketsToTransform.metadata.continuousStats?.copy(lastTimestamp = Instant.now())
```
thalurur, line 182: Just curious: What is this update for? I thought we always compute the continuous stats when explain API is called, why are we storing this?

src/main/kotlin/org/opensearch/indexmanagement/transform/TransformSearchService.kt
```diff
@@ -119,15 +232,45 @@ class TransformSearchService(
                 .allowPartialSearchResults(false)
         }
 
-        fun convertResponse(transform: Transform, searchResponse: SearchResponse, waterMarkDocuments: Boolean = true): TransformSearchResult {
+        fun getBucketSearchRequest(
```
thalurur, line 235: minor: rename to `getShardLeveBucketSearchRequest`?

src/main/kotlin/org/opensearch/indexmanagement/transform/TransformSearchService.kt
```diff
@@ -65,8 +79,71 @@ class TransformSearchService(
...
+            throw TransformSearchServiceException(getShardsErrorMessage, unwrappedException)
+        } catch (e: OpenSearchSecurityException) {
+            throw TransformSearchServiceException("$getShardsErrorMessage - missing required index permissions: ${e.localizedMessage}", e)
+        } catch (e: Exception) {
+            throw TransformSearchServiceException(getShardsErrorMessage, e)
+        }
+    }
+
+    @Suppress("RethrowCaughtException")
+    suspend fun getModifiedBuckets(transform: Transform, afterKey: Map<String, Any>?, currentShard: ShardNewDocuments): BucketSearchResult {
```
thalurur, line 111: minor: rename `getShardLevelModifiedBuckets` or `getModifiedBucketsInShard`?

src/main/kotlin/org/opensearch/indexmanagement/transform/TransformSearchService.kt
```diff
@@ -144,6 +287,14 @@ class TransformSearchService(
             return TransformSearchResult(stats, docsToIndex, afterKey)
         }
 
+        // Gathers and returns from the bucket search response the modified buckets from the query, the afterkey, and the search time
+        private fun convertBucketSearchResponse(transform: Transform, searchResponse: SearchResponse): BucketSearchResult {
+            val aggs = searchResponse.aggregations.get(transform.id) as CompositeAggregation
+            val bucketSearchResult = BucketSearchResult(HashSet(), aggs.afterKey(), searchResponse.took.millis)
+            bucketSearchResult.modifiedBuckets.addAll(aggs.buckets.map { it.key })
```
thalurur, line 294: why not just ```val bucketSearchResult = BucketSearchResult(aggs.buckets.map { it.key }.toSet(), aggs.afterKey(), searchResponse.took.millis)```?

src/main/kotlin/org/opensearch/indexmanagement/transform/TransformRunner.kt
```diff
@@ -140,46 +157,123 @@ object TransformRunner :
...
+        // If we have not populated the list of shards to search, do so now
+        if (bucketsToTransform.shardsToSearch == null) {
+            val oldShardIDToMaxSeqNo = bucketsToTransform.metadata.shardIDToGlobalCheckpoint
+            val shardIdToGlobalCheckpoint = withTransformSecurityContext(transform) {
+                transformSearchService.getShardsGlobalCheckpoint(transform.sourceIndex)
+            }
+            val shardsToSearch = getShardsToSearch(oldShardIDToMaxSeqNo, shardIdToGlobalCheckpoint).iterator()
+            currentBucketsToTransform = bucketsToTransform.copy(
+                metadata = bucketsToTransform.metadata.copy(
+                    shardIDToGlobalCheckpoint = shardIdToGlobalCheckpoint,
```
thalurur, line 181: is my understanding correct - we update the shardIdToGlobalCheckpoint in metadata before we actually transform new delta. i.e if the node on which the transform is running dies after metadata update we can potentially not transform the data, but show to the user we did transform in the explain?

src/main/kotlin/org/opensearch/indexmanagement/transform/TransformSearchService.kt
downsrob: For non continuous transforms, we set missingBucket to True, which means that there will be a bucket with a null key created the represent docs which have that field missing. 
I think for continuous transforms, the null fields could be searched for with a 'must not exist' query. Do you think we should include the missing buckets for continuous transforms, as it is for non continuous? 
I believe that the reason I opted to exclude it was that if there are some random documents which come in with null values in the grouping fields, then this could result in some very large buckets with the null key being rerun every transform iteration, when the user may not even want those buckets. We could also expose missingBucket as a setting for any transform in the future if needed.

src/main/kotlin/org/opensearch/indexmanagement/transform/TransformRunner.kt
downsrob: At this point we are just storing the lastTimestamp, as that represents the time when the transform got the shard global checkpoints, so a user might know the time boundaries around what data is included in the transform iteration

src/main/kotlin/org/opensearch/indexmanagement/transform/TransformRunner.kt
downsrob: You are right, shardIdToGlobalCheckpoint would be updated with the first successful transform iteration, after indexing the first page of the transform's actual composite search. At this point, any failure would stop the transform and put it into a FAILED state, but in an explain response, the continuous stats would reflect a success. I don't believe that we support retrying transforms, so I don't think that next runs are a problem, but the continuous stats response may be confusing.

Not updating the stats until the end may not be correct either, though, as the buckets must be partially indexed at this point. Also, if a user is doing a bulk explain on their transforms, then the user may not care for the dynamic calculation of the documents behind for the failed transform, so I would prefer to not return continuous stats at all for failed transforms. 
What do you think?

downsrob: Nevermind, we do support retrying failed transforms through the start transform api, so to avoid missed data, either shardIdToGlobalCheckpoint should not be updated until the end of an execution, or we need to clear it on retry. I will try to move updating it to the end of the execution.

src/main/kotlin/org/opensearch/indexmanagement/transform/action/explain/TransportExplainTransformAction.kt
downsrob: I added individual failures for failures during parsing the metadata response or while adding continuous transform stats. I don't see many possible failures besides for getting the continuous transform stats, but if you think it is worth adding transform-level failures for all failures in this transport action, let me know and I will add that.

src/main/kotlin/org/opensearch/indexmanagement/transform/TransformSearchService.kt
thalurur: Slightly prefer doing the null by default and skipping with a setting later on, but upto you. If we skip null buckets lets call it out in documentation for continuous transforms

src/main/kotlin/org/opensearch/indexmanagement/transform/action/explain/TransportExplainTransformAction.kt
thalurur: I think this should be a two way door, lets keep it as it is and if we see issues we can revisit to make the failure granular

src/main/kotlin/org/opensearch/indexmanagement/transform/model/ContinuousTransformStats.kt
```diff
@@ -0,0 +1,80 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
```
thalurur, line 10: change the license?

```
/*
 * Copyright OpenSearch Contributors
 * SPDX-License-Identifier: Apache-2.0
 */
```

src/main/kotlin/org/opensearch/indexmanagement/transform/action/explain/TransportExplainTransformAction.kt
```diff
@@ -129,6 +146,25 @@ class TransportExplainTransformAction @Inject constructor(
...
+                                        idsToExplain.computeIfPresent(metadata.transformId) { _, explainTransform ->
+                                            explainTransform.copy(
+                                                metadata = metadata.copy(
+                                                    shardIDToGlobalCheckpoint = null,
+                                                    continuousStats = continuousStats
+                                                )
+                                            )
+                                        }
+                                    }
+                                }
```
thalurur, line 167: nit: can be moved to transform class?

src/main/kotlin/org/opensearch/indexmanagement/transform/TransformSearchService.kt
downsrob: Added null by default. I do think that it may be worthwhile someday to expose missingBucket as a transform setting someday in general though, it may be useful for non continuous transforms as well. I added an integration test as well, it ended up being pretty verbose so let me know if you have an improvement.

src/main/kotlin/org/opensearch/indexmanagement/transform/model/ContinuousTransformStats.kt
downsrob: Good catch, thanks!

thalurur (approved):

dbbaughe (approved):

alexmaurizio:
> > Do you want this behavior? Otherwise you could just update the IndexTransform transport action and reject any request that tries to change the continuous flag (we do it for rollup).
> 
> I don't see any use case for changing a transform from continuous to non-continuous, and I could go either way on non-continuous to continuous, so I would opt for handling this like rollups, just to maintain consistency. It would be an easy change later on anyways

Hello, after the (long awaited) release of OS 1.3 on the AWS Service, I'd just like to point out that -there were- a lot of uses case for the upgrade from Non-Continuous to Continuous. I understand the first time it had to be re-ran, but after that the correct checkpointing data would be added.

Now, I have more than 30 transforms running in production on billions of documents and I ran them manually every other week. Without the ability to upgrade them, I have to manually rebuild all of the transforms with new IDs, point them to the same target index, then surgically cut the timestamp data to match the transforms with delete_by_queries on the index with the two different transform IDs. And I have one of the transforms that do NOT use a timestamp field, so I'm both confused on how to approach it and how will continuous handle that specific case without repassing all the data everytime it runs.

Keep up the good work anyway :)

Issue #173, opened 2021-10-27
Add MacOS Support


dblock:
Coming from https://github.com/opensearch-project/opensearch-plugins/issues/96, add MacOS (Darwin) support.

- [ ] Passing CI on MacOS
- [ ] Documentation

Issue #172, opened 2021-10-27
Add Windows Support


dblock:
Coming from https://github.com/opensearch-project/opensearch-plugins/issues/95, add Windows support.

- [ ] Passing CI on Windows
- [ ] Documentation

PR #210, opened 2021-12-03
Adds test and build workflow for mac and windows

downsrob:
*Issue #, if available:*
https://github.com/opensearch-project/index-management/issues/173 https://github.com/opensearch-project/index-management/issues/172
*Description of changes:*

- Adds windows and mac build tasks to the test and build workflow to run as part of github CI. As there currently are no min distributions for mac or windows, plugins are blocked on running integration tests on mac or windows machines. While waiting for the min distribution, integration tests are disabled, but unit tests and build tasks are run. 
- Artifacts are uploaded for each build. Though the artifacts differ very little, with just different line ending for windows, and different manifests documenting the build OS, they may be useful for debugging a build failure on the mac or windows distribution.

*CheckList:*

### DIFF
--- a/.github/workflows/test-and-build-workflow.yml
+++ b/.github/workflows/test-and-build-workflow.yml
@@ -65,7 +65,7 @@ jobs:
         run: subst 'X:' .
       - name: Build with Gradle
         working-directory: 'X:'
-        run: ./gradlew.bat build -D"opensearch.version=1.2.0-SNAPSHOT" -x integTest -x jacocoTestReport
+        run: ./gradlew.bat build -D"opensearch.version=1.3.0-SNAPSHOT" -x integTest -x jacocoTestReport
         env:
           _JAVA_OPTIONS: -Xmx4096M
 #      - name: Upload failed logs
@@ -99,7 +99,7 @@ jobs:
       - name: Checkout Branch
         uses: actions/checkout@v2
       - name: Build with Gradle
-        run: ./gradlew build -Dopensearch.version=1.2.0-SNAPSHOT -x integTest -x jacocoTestReport
+        run: ./gradlew build -Dopensearch.version=1.3.0-SNAPSHOT -x integTest -x jacocoTestReport
 #      - name: Upload failed logs
 #        uses: actions/upload-artifact@v2
 #        if: failure()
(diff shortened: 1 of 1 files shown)


dbbaughe:
@downsrob Do we also need to add windows/mac to the multi node workflow?

downsrob:
> @downsrob Do we also need to add windows/mac to the multi node workflow?

@dbbaughe We may want to at a later stage, but the multi node workflow only runs the integTest gradle action, which is also the only action which doesn't work because of missing the min distribution. So there is no point adding it in now I think

dbbaughe:
@dblock

As we add these matrix of tests (OS', JVMs, BWC, Perf, Plugin specifc, etc.) we're going to start seeing plugins running a huge amount of tests on the runners for each repository.

Are we worried at all about GitHub runner limits? Are there any limits for our org for how many can be run in parallel at the same time? Definitely don't want to start seeing these get queue'd up and take hours to finish for simple PRs.

.github/workflows/test-and-build-workflow.yml
```diff
@@ -43,5 +43,76 @@ jobs:
...
         uses: actions/upload-artifact@v1
         with:
-          name: index-management-plugin
+          name: index-management-plugin-ubuntu
           path: index-management-artifacts
+  windows-build:
+    # Job name
+    name: Build Index Management
+    # This job runs on Windows
+    runs-on: windows-latest
```
dbbaughe, line 52: Is it not possible to use the matrix for the runs-on so the steps aren't duplicated each time?

dbbaughe (dismissed):

.github/workflows/test-and-build-workflow.yml
```diff
@@ -1,5 +1,5 @@
 name: Test and Build Workflow
-# This workflow is triggered on pull requests to main or a opendistro release branch
+# This workflow is triggered on pull requests to main or an opendistro release branch
```
thalurur, line 2: nit remove?

thalurur (dismissed):

thalurur (approved):

dbbaughe (approved):

PR #211, opened 2021-12-03
Fixes Flakey Tests

downsrob:

*Issue #, if available:*
https://github.com/opensearch-project/index-management/issues/90
*Description of changes:*

- Running multi node integTest on a t2.large EC2 instance would fail around 10% of the time before this change, this was tested using a shell script to run the integration test 100 times.
- The majority of failures were due to jobs not starting. This appeared to be due to the job scheduler not setting the index operation listener on the index before the force start was attempted. These failures would often say "metadata document not created" or "target index not created".
- By initializing the job index before any tests are run, the job sweeper can register a listener on the job index before the job is created, and catch the force start when it occurs. The github hosted runners are slightly lower resource than the t2.large, and they have additional background tasks, so it is possible that some flakiness could persist here, if that is the case, sleeping for a few seconds after initializing the index should help.
- Fixes the 'test disabling and reenabling ism' test flakiness by allowing 4 retryOnConflicts. It seems that in all of the forcing starts and updating cluster settings, that sometimes the managed index config would be modified while forcing the start. This additional option allows for retrying the force start if it fails. In my tests, this removed all flakiness.

*CheckList:*

### DIFF
--- a/src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
+++ b/src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
@@ -14,7 +14,9 @@ import org.opensearch.client.Request
 import org.opensearch.client.RequestOptions
 import org.opensearch.client.Response
 import org.opensearch.client.RestClient
+import org.opensearch.common.Strings
 import org.opensearch.common.settings.Settings
+import org.opensearch.indexmanagement.indexstatemanagement.util.INDEX_HIDDEN
 import org.opensearch.rest.RestStatus
 import java.nio.file.Files
 import java.nio.file.Path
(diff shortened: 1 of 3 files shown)


dbbaughe:
Any insights around the 'test disabling and reenabling ism' test?

dbbaughe (approved):

downsrob:
> Any insights around the 'test disabling and reenabling ism' test?

Not yet, the failure is: 
`{"error":{"root_cause":[{"type":"version_conflict_engine_exception","reason":"[5kpYRYqfQqCycTwvkqxqhA]: version conflict, required seqNo [7], primary term [1]. current document has seqNo [8] and primary term [1]", "index":".opendistro-ism-config","shard":"0","index_uuid":"rrtpcbBuR0CGImHh8rNATQ"},"status":409}`
On my EC2 instances I can only replicate this every 1 or 2/100 attempts, so it is much slower to debug, but I am going to add additional debugging around this case and see if I can identify what the intermediate operation is that is causing the sequence number to be out of date.

dbbaughe:
@downsrob Are you able to replicate it every 1 or 2/100 attempts even when it's the only test that is run? Should be able to run that test 100 times in succession a lot faster than the entire test suite. If not perhaps it has to do with another test bleeding over somehow.

downsrob:
@dbbaughe That is a good idea, I started the single test run to see. If that doesn't work then I can try the test suite. Using a failing test seed might be important too, as these flakey failures often depend on test order. 

Do you know how the test seeds are applied? I have often wondered if the test seed is set per test, per suite, or per overall run. I am curious, as if the seed is set once at the start of the overall run, then it may not guarantee the same test order or values when only the specific tests or test suites are run.

dbbaughe (approved):

thalurur (approved):

Issue #191, opened 2021-10-15
Index age tracking since rollover


Lawryy:
Index state management using min_index_age doesn't guarantee a definite amount of retention time when using size-based index rollovers.

E.g. Data retention has to be 30d or more:

Index rollover happens once per week, when shard size hits 30GB. Then it moves to delete phase with min_index_age set to 30d. Index gets deleted after 30 days its creation, but since logs were still going in a week after its creation, the newest messages that get deleted are less than 30 days old. 

Add a min_age_since_rollover (or similar) variable that can be used with Index State Management.

Or add a variable to "lock" the index in a state for a defined amount of time, to ensure sufficient retention.

ILM with Elasticsearch already behaves this way with rollovers

dbbaughe:
@Lawryy

We should be able to get the time a rollover happened for an index from the metadata.
That being said, an index can have multiple aliases that were rolled over on it:
e.g.
Create foo-index w/ alias bar and alias baz

You can call /bar/_rollover/new-index-1 and /baz/_rollover/another-index-1
And now the metadata for foo-index will show that it had two rollovers, one for bar and another for baz each with different timestamps.

We do not want to have concrete alias names in the policy itself to know which to use as that would mean it's not reusable across multiple groups of indices potentially.

So what do you think about:

Introducing min_age_since_rollover condition where you can specify a time unit like 10h, 5d, 30d, etc. and by default it'll use the _oldest_ rollover it finds for that index (the most common case will probably only have a single rollover).

If it finds no rollover has happened... need to decide if it should continue waiting for a rollover to happen or if it should fail because to use this condition check a rollover **must** have happened and no rollover is an error.

And then we could introduce a new optional index setting to let you explicitly say which alias to use for the check for a specific index if for some reason you have multiple rollovers on an index.

Lawryy:
Using the oldest rollover sounds fine to me. 

I think it would be wise to give an error if there is no rollover, as the check should only be applied after the rollover has happened. 

The optional index setting also seems like a good idea.

Thanks!

PR #215, opened 2021-12-07
Adds min rollover age as a transition condition

dbbaughe:

*Issue #, if available:*
https://github.com/opensearch-project/index-management/issues/191

*Description of changes:*

Adds a new transition condition - min_rollover_age

This lets you transition to a new state based on when the index was rolled over instead of the index creation date to help with keeping data around for a specific amount of time.

As an index can be rolled over multiple times using different aliases (which is uncommon, i.e. I haven't actually seen that use case yet, so if you have this use case please comment on the PR) we will use the oldest rollover date that is available, i.e. the first one that happened. In the future if it seems like users want to specify which alias to use for this calculation we can introduce an index level setting to allow specifying the exact alias to use. I didn't include it for now in this PR as it seems like an unknown use case and we're starting to have a lot of settings introduced so want to limit them.

If the index evaluates this condition and the index was never rolled over, we will fail as this transition condition is only meant to be used after a rollover action has occured.

*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/Transition.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/Transition.kt
@@ -99,11 +99,12 @@ data class Conditions(
     val indexAge: TimeValue? = null,
     val docCount: Long? = null,
     val size: ByteSizeValue? = null,
-    val cron: CronSchedule? = null
+    val cron: CronSchedule? = null,
+    val rolloverAge: TimeValue? = null
 ) : ToXContentObject, Writeable {
 
     init {
-        val conditionsList = listOf(indexAge, docCount, size, cron)
+        val conditionsList = listOf(indexAge, docCount, size, cron, rolloverAge)
         require(conditionsList.filterNotNull().size == 1) { "Cannot provide more than one Transition condition" }
 
         // Validate doc count condition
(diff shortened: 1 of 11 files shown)


thalurur (approved):

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/Transition.kt
```diff
@@ -162,11 +168,12 @@ data class Conditions(
                     MIN_DOC_COUNT_FIELD -> docCount = xcp.longValue()
                     MIN_SIZE_FIELD -> size = ByteSizeValue.parseBytesSizeValue(xcp.text(), MIN_SIZE_FIELD)
                     CRON_FIELD -> cron = ScheduleParser.parse(xcp) as? CronSchedule
+                    MIN_ROLLOVER_AGE_FIELD -> rolloverAge = TimeValue.parseTimeValue(xcp.text(), MIN_ROLLOVER_AGE_FIELD)
```
downsrob, line 171: I think all range checking for this field is being done by the TimeValue parser, but it looks like TimeValue allows for magic numbers -1 and 0. Looking through the code, I don't think that this will cause any trouble, but you might want to test it, as it seems like we might allow -1 or 0 from users.
https://github.com/opensearch-project/OpenSearch/blob/main/libs/core/src/main/java/org/opensearch/common/unit/TimeValue.java#L392-L395

downsrob (approved):

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/Transition.kt
dbbaughe: Yeah, they just end up being their value, i.e. 0 -> 0 and -1 -> -1 when converting to millis and for the condition checks they'll essentially just always evaluate to true. Not sure anyone would actually use it (they can just not use a condition), but will keep it as is since it's the same as index_age too right now.

Issue #213, opened 2021-11-09
Add max_primary_shard_size as a rollover action in State Management Policy


Kevc1:
**Is your feature request related to a problem? Please describe.**
We need a way to rollover our indices if the shard size is >50gb as recommended by AWS.  This option exists in Elasticsearch and would be very beneficial to have available in Opensearch.

**Describe the solution you'd like**
Would like to have this option available in an ISM Policy, such as belwo:
                "name": "hot",
                "actions": [
                    {
                        "rollover": {
                            "min_size": "90gb",
                            "min_index_age": "7d",
                            "max_primary_shard_size" : "50gb"

                        }
                     },
                  ]

**Describe alternatives you've considered**

**Additional context**

dbbaughe:
@Kevc1 

What is the expectation in terms of evaluating the condition to true:
* When at least one of the primary shards is > 50GB
* When the majority of primary shards are > 50GB
* When all of the primary shards are > 50GB

I'm guessing it's the first one.. but do you see value in the others at all?

Kevc1:
I was initially thinking the second one but if any shard is going above the limit we probably want to rollover, so #1 would probebly be the safest.

dbbaughe:
@Kevc1, Yes, the second and third one could end up with a hot shards issue where a lot of data is written to specific shards if using something like custom routing, perhaps before the rollover condition is met you have some shards that are already hundreds of GB or more.

dbbaughe:
Merged into main, will go out in the next release (1.3).

PR #220, opened 2021-12-09
Adds min primary shard size rollover condition to the ISM rollover ac…

dbbaughe:
…tion

*Issue #, if available:*
https://github.com/opensearch-project/index-management/issues/213

*Description of changes:*
* Adds min_primary_shard_size as a rollover condition to ISM rollover action.
* It will evaluate to true when there is a single primary shard in the index that has a greater size than the condition.
* Kept the prefix as min instead of max to be aligned with the other conditions we have in rollover, although it is odd that our conditions are differently named than the native rollover API conditions.

*CheckList:*

### DIFF
--- a/src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
+++ b/src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
@@ -29,7 +29,7 @@ import javax.management.remote.JMXServiceURL
 
 abstract class IndexManagementRestTestCase : ODFERestTestCase() {
 
-    val configSchemaVersion = 12
+    val configSchemaVersion = 13
     val historySchemaVersion = 3
 
     // Having issues with tests leaking into other tests and mappings being incorrect and they are not caught by any pending task wait check as
--- a/src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
+++ b/src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
@@ -13,11 +13,13 @@ import org.junit.rules.DisableOnDebug
 import org.opensearch.client.Request
 import org.opensearch.client.RequestOptions
 import org.opensearch.client.Response
+import org.opensearch.client.ResponseListener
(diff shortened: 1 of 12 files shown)


dbbaughe:
TODO: Open a PR on dashboards plugin to add this new option to the UI
This isn't blocking to get this PR in though for now

qreshi (dismissed):

src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
```diff
@@ -106,11 +108,20 @@ abstract class IndexManagementRestTestCase : ODFERestTestCase() {
...
-    protected fun insertSampleData(index: String, docCount: Int, delay: Long = 0, jsonString: String = "{ \"test_field\": \"test_value\" }") {
+    protected fun insertSampleData(index: String, docCount: Int, delay: Long = 0, jsonString: String = "{ \"test_field\": \"test_value\" }", routing: String? = null) {
+        var endpoint = "/$index/_doc/?refresh=true"
+        if (routing != null) endpoint += "&routing=$routing"
         for (i in 1..docCount) {
-            val request = Request("POST", "/$index/_doc/?refresh=true")
+            val request = Request("POST", endpoint)
             request.setJsonEntity(jsonString)
-            client().performRequest(request)
+            client().performRequestAsync(
```
downsrob, line 117: Do you see this as potentially causing flakiness in the future? It seems to me like the test could continue on before all of the sample data is available.

dbbaughe: Could definitely happen, was leftover when I was previously increasing shard size by adding more and more documents and this test function was taking a long time. Changed to increasing the payload size instead which doesn't have the issue, reverted.

downsrob (approved):

qreshi (approved):

sastorsl:
Should perhaps have updated the docs aswell: [https://opensearch.org/docs/latest/im-plugin/ism/policies/#rollover](https://opensearch.org/docs/latest/im-plugin/ism/policies/#rollover)

PR #238, opened 2021-12-27
Refactors Notification action with new interface

downsrob:
*Issue #, if available:*

*Description of changes:*
Refactors Notification action with new interface. To support the settings dynamically used in the notification action, as well as the script service used in the notification and transition actions, the settings and script service fields were added to the step context.

*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
@@ -13,7 +13,6 @@ import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import org.apache.logging.log4j.LogManager
-import org.locationtech.jts.algorithm.Orientation.index
 import org.opensearch.action.admin.cluster.state.ClusterStateRequest
 import org.opensearch.action.admin.cluster.state.ClusterStateResponse
 import org.opensearch.action.bulk.BackoffPolicy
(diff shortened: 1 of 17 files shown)


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/notification/AttemptNotificationStep.kt
```diff
@@ -5,24 +5,76 @@
...
+    private var stepStatus = StepStatus.STARTING
+    private var info: Map<String, Any>? = null
 
     override suspend fun execute(): Step {
-        TODO("Not yet implemented")
+        if (this.context == null || this.context?.settings == null || this.context?.scriptService == null) return this
+        val context = this.context!!
+        val indexName = context.metadata.index
+        val hostDenyList = context.settings!!.getAsList(ManagedIndexSettings.HOST_DENY_LIST)
+        val scriptService = context.scriptService!!
```
thalurur, line 32: We don't have to have the !! assertion in the code, I think if you save the context to a local variable and then do a null check on top of it the compiler will not complain about it being null

e.g
```
val context == this.context 
if (context == null || context.settings == null || context.scriptService == null) return this
```

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -267,7 +266,7 @@ object ManagedIndexRunner :
 
         val state = policy.getStateToExecute(managedIndexMetaData)
         val action: Action? = state?.getActionToExecute(managedIndexMetaData.copy(user = policy.user, threadContext = threadPool.threadContext))
-        val stepContext = StepContext(managedIndexMetaData, clusterService, client, null, null)
+        val stepContext = StepContext(managedIndexMetaData, clusterService, client, null, null, scriptService, settings)
```
thalurur, line 269: it probably is the case we are going to populate the script service and settings always in the runner - do you think we can may be make these non nullable in the StepContext or is any other benefit to having those nullable?

downsrob: I agree, they should not have been nullable. My original thinking was that the parameters should be nullable as they are only used in one or two actions, but we are always populating the step context from the runner so there is no point.

thalurur (dismissed):

bowenlan-amzn (approved):

leeyun-amzn (approved):

PR #240, opened 2021-12-30
Porting missing logic 

thalurur:

*Issue #, if available:*
N/A

*Description of changes:*
* Support writing to history indices when they are configured as protected indices 
* Always returning total managed indices count in ISM explain API 
* Porting missing changes - using scroll API to sweep managed index for managed jobs

*CheckList:*


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/util/ManagedIndexUtils.kt
```diff
@@ -175,48 +173,29 @@ fun updateManagedIndexRequest(sweptManagedIndexConfig: SweptManagedIndexConfig):
...
-                .fetchSource(
-                    arrayOf(
-                        "${ManagedIndexConfig.MANAGED_INDEX_TYPE}.${ManagedIndexConfig.INDEX_FIELD}",
-                        "${ManagedIndexConfig.MANAGED_INDEX_TYPE}.${ManagedIndexConfig.INDEX_UUID_FIELD}",
-                        "${ManagedIndexConfig.MANAGED_INDEX_TYPE}.${ManagedIndexConfig.POLICY_ID_FIELD}",
-                        "${ManagedIndexConfig.MANAGED_INDEX_TYPE}.${ManagedIndexConfig.CHANGE_POLICY_FIELD}"
-                    ),
-                    emptyArray()
-                )
+                .fetchSource(emptyArray(), emptyArray())
```
dbbaughe, line 198: What's this doing now? Does it get the whole document or passing emptyArrays means it gets nothing?

thalurur: Before we were fetching these to create the `SweptManagedIndexConfig` data model object but we don't really use any of the fields we are fetching through out the call chain, rather just use the managed index uuid so refactored to not fetch the source at all

dbbaughe (dismissed):

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
```diff
@@ -577,66 +576,83 @@ class ManagedIndexCoordinator(
...
 
-        updateManagedIndices(
-            updateMatchingIndicesReqs + deleteManagedIndexRequests,
-            updateMatchingIndicesReqs.isNotEmpty()
-        )
+        updateManagedIndices(updateMatchingIndicesReqs + deleteManagedIndexRequests)
 
         // clean metadata of un-managed index
-        val indicesToDeleteMetadataFrom =
-            unManagedIndices + getManagedIndicesToDelete(currentIndices, currentManagedIndices)
```
bowenlan-amzn, line 603: Not sure why `getManagedIndicesToDelete` is removed.

thalurur: good catch, this can cause dangling metadata docs as if the managed index is deleted in a sweep and current index doesn't have that index anymore then the metadata doc will never be deleted otherwise. Will fix

bowenlan-amzn (approved):

dbbaughe (approved):

PR #245, opened 2022-01-04
Makes rest APIs use new metadata service

downsrob:
*Issue #, if available:*

*Description of changes:*
**SPI**
- Makes IndexMetadataService.getMetadata into a suspend method
- Adds indexCreationDate to ManagedIndexMetadata
- Adds loadExtensions code to populate the list of action parsers and metadata providers from the plugin extensions
- Creates the DefaultIndexMetadataService to use a cluster state rest call to get a map of index names to index uuid and creation date given a list of index names
- Adds IndexMetadataProvider class to manage the map of index types to metadata services
  - if an IndexMetadataService is added for an index type which is already mapped, an error is thrown, the cluster would fail to start
- Adds getIndexMetadataWriteOverrideSettings to the IndexMetadataProvider class to return the list of non-null IndexMetadataWriteOverrideSettings from the IndexMetadataServices. This is used in the Update Managed Index Metadata rest api

**Add Policy**
- Makes add policy rest API use the default index metadata service provided by the IndexMetadataProvider.
- As the metadata service resolves index names, the index name resolver was removed from the add policy transport
- Following that, validating index permissions needed to occur before other validations, as we don't want to include the resolved index name of an index which the user does not have permissions to view in the failedIndices response

**Explain**
- Combines ExplainAll into Explain, making totalManagedIndices and enabledState included in all explain responses.
- Adds typealias and constants to the TransportExplainAction class
- Uses indexMetadataProvider to resolve index names and get index names and types from the provider specific to the index type

**Change Policy**
- Uses indexMetadataProvider to resolve index names and get index names and types from the provider specific to the index type
- Gets the cluster state for default index types to check if metadata migration is ongoing

**Remove Policy**
- Uses indexMetadataProvider to resolve index names and get index names and types from the provider specific to the index type
- Gets the cluster state for default index types to fail closed indices, and set auto manage to false on indices before removing the policy

**Retry Failed Managed Index**
- Uses indexMetadataProvider to resolve index names and get index names and types from the provider specific to the index type
- Gets the cluster state for default index types to check if metadata migration is ongoing

**Get Policy/Policies**
- No refactoring changes needed
- Added simple integration tests

**Delete Policy**
- No refactoring changes needed
- Added simple integration test

**Update Managed Index Metadata**
- Uses IndexMetadataWriteOverrideSetting from each index metadata provider to enable extensions to have the option of overriding the cluster metadata write block. If any extension provides an index setting for the IndexMetadataWriteOverrideSetting which resolves to true, the metadata write will be allowed for that index, despite the cluster metadata write block.

*CheckList:*

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -306,9 +306,6 @@ integTest {
     if (System.getProperty("tests.clustername") != null) {
         exclude 'org/opensearch/indexmanagement/indexstatemanagement/MetadataRegressionIT.class'
     }
-
-    // TODO: Remove me after refactoring all actions
-    exclude 'org/opensearch/indexmanagement/indexstatemanagement/resthandler/*'
 }
 
 String bwcVersion = "1.13.2.0"
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/IndexMetadataProvider.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/IndexMetadataProvider.kt
@@ -42,7 +42,13 @@ class IndexMetadataProvider(
 
     suspend fun getISMIndexMetadataByType(type: String = DEFAULT_INDEX_TYPE, indexNames: List<String>): Map<String, ISMIndexMetadata> {
         val service = services[type] ?: throw IllegalArgumentException(getTypeNotRecognizedMessage(type))
+        if (type != DEFAULT_INDEX_TYPE && indexNames.size > 1) throw IllegalArgumentException(MULTIPLE_INDICES_CUSTOM_INDEX_TYPE_ERROR)
         return service.getMetadata(indexNames, client, clusterService)
+    }
+
+    suspend fun getAllISMIndexMetadataByType(type: String = DEFAULT_INDEX_TYPE): Map<String, ISMIndexMetadata> {
+        val service = services[type] ?: throw IllegalArgumentException(getTypeNotRecognizedMessage(type))
+        return service.getMetadataForAllIndices(client, clusterService)
     }
 
     /*
@@ -53,6 +59,7 @@ class IndexMetadataProvider(
         types: List<String> = services.keys.toList(),
         indexNames: List<String>
     ): Map<String, Map<String, ISMIndexMetadata>> = coroutineScope {
+        if (types.any { it != DEFAULT_INDEX_TYPE } && indexNames.size > 1) throw IllegalArgumentException(MULTIPLE_INDICES_CUSTOM_INDEX_TYPE_ERROR)
         val requests = ArrayList<Deferred<Pair<String, Map<String, ISMIndexMetadata>>>>()
         // Start all index metadata requests at the same time
         types.forEach { type ->
@@ -82,8 +89,13 @@ class IndexMetadataProvider(
         metadata
     }
 
+    fun getIndexMetadataWriteOverrideSettings(): List<String> {
+        return services.values.mapNotNull { it.getIndexMetadataWriteOverrideSetting() }
+    }
+
     companion object {
         const val EVALUATION_FAILURE_MESSAGE = "Matches restricted index pattern defined in the cluster setting"
+        const val MULTIPLE_INDICES_CUSTOM_INDEX_TYPE_ERROR = "Cannot get metadata for more than one index name/pattern when using a custom index type"
         fun getTypeNotRecognizedMessage(indexType: String) = "Index type [type=$indexType] was not recognized when trying to get index metadata"
         fun getDuplicateServicesMessage(indexType: String) = "Multiple metadata services attempted to assign a service to the index type [$indexType]"
     }
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
@@ -304,11 +304,8 @@ object ManagedIndexRunner :
         }
 
         val state = policy.getStateToExecute(managedIndexMetaData)
-        val action: Action? = state?.getActionToExecute(
-            managedIndexMetaData.copy(user = policy.user, threadContext = threadPool.threadContext),
-            indexMetadataProvider
-        )
-        val stepContext = StepContext(managedIndexMetaData, clusterService, client, null, null, scriptService, settings)
+        val action: Action? = state?.getActionToExecute(managedIndexMetaData, indexMetadataProvider)
+        val stepContext = StepContext(managedIndexMetaData, clusterService, client, threadPool.threadContext, policy.user, scriptService, settings)
         val step: Step? = action?.getStepToExecute(stepContext)
         val currentActionMetaData = action?.getUpdatedActionMetadata(managedIndexMetaData, state.name)
 
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/AddPolicyRequest.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/AddPolicyRequest.kt
@@ -10,23 +10,31 @@ import org.opensearch.action.ActionRequestValidationException
 import org.opensearch.action.ValidateActions
 import org.opensearch.common.io.stream.StreamInput
 import org.opensearch.common.io.stream.StreamOutput
+import org.opensearch.indexmanagement.indexstatemanagement.util.DEFAULT_INDEX_TYPE
 import java.io.IOException
 
 class AddPolicyRequest(
     val indices: List<String>,
-    val policyID: String
+    val policyID: String,
+    val indexType: String
 ) : ActionRequest() {
 
     @Throws(IOException::class)
     constructor(sin: StreamInput) : this(
         indices = sin.readStringList(),
-        policyID = sin.readString()
+        policyID = sin.readString(),
+        indexType = sin.readString()
     )
 
     override fun validate(): ActionRequestValidationException? {
         var validationException: ActionRequestValidationException? = null
         if (indices.isEmpty()) {
             validationException = ValidateActions.addValidationError("Missing indices", validationException)
+        } else if (indexType != DEFAULT_INDEX_TYPE && indices.size > 1) {
+            validationException = ValidateActions.addValidationError(
+                MULTIPLE_INDICES_CUSTOM_INDEX_TYPE_ERROR,
+                validationException
+            )
         }
         return validationException
     }
@@ -35,5 +43,11 @@ class AddPolicyRequest(
     override fun writeTo(out: StreamOutput) {
         out.writeStringCollection(indices)
         out.writeString(policyID)
+        out.writeString(indexType)
+    }
+
+    companion object {
+        const val MULTIPLE_INDICES_CUSTOM_INDEX_TYPE_ERROR =
+            "Cannot add policy to more than one index name/pattern when using a custom index type"
     }
 }
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
@@ -5,21 +5,9 @@
 
 package org.opensearch.indexmanagement.indexstatemanagement.transport.action.addpolicy
 
-/*
- * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
- *
- * Licensed under the Apache License, Version 2.0 (the "License").
- * You may not use this file except in compliance with the License.
- * A copy of the License is located at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
- * or in the "license" file accompanying this file. This file is distributed
- * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
- * express or implied. See the License for the specific language governing
- * permissions and limitations under the License.
- */
-
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
 import org.apache.logging.log4j.LogManager
 import org.opensearch.ExceptionsHelper
 import org.opensearch.OpenSearchSecurityException
@@ -39,9 +27,7 @@ import org.opensearch.action.support.HandledTransportAction
 import org.opensearch.action.support.IndicesOptions
 import org.opensearch.action.support.master.AcknowledgedResponse
 import org.opensearch.client.node.NodeClient
-import org.opensearch.cluster.ClusterState
 import org.opensearch.cluster.block.ClusterBlockException
-import org.opensearch.cluster.metadata.IndexNameExpressionResolver
 import org.opensearch.cluster.service.ClusterService
 import org.opensearch.common.inject.Inject
 import org.opensearch.common.settings.Settings
@@ -49,7 +35,9 @@ import org.opensearch.common.unit.TimeValue
 import org.opensearch.common.xcontent.NamedXContentRegistry
 import org.opensearch.commons.ConfigConstants
 import org.opensearch.commons.authuser.User
+import org.opensearch.index.Index
 import org.opensearch.indexmanagement.IndexManagementPlugin.Companion.INDEX_MANAGEMENT_INDEX
+import org.opensearch.indexmanagement.indexstatemanagement.DefaultIndexMetadataService
 import org.opensearch.indexmanagement.indexstatemanagement.IndexMetadataProvider
 import org.opensearch.indexmanagement.indexstatemanagement.IndexMetadataProvider.Companion.EVALUATION_FAILURE_MESSAGE
 import org.opensearch.indexmanagement.indexstatemanagement.model.Policy
@@ -58,10 +46,13 @@ import org.opensearch.indexmanagement.indexstatemanagement.settings.ManagedIndex
 import org.opensearch.indexmanagement.indexstatemanagement.transport.action.ISMStatusResponse
 import org.opensearch.indexmanagement.indexstatemanagement.transport.action.managedIndex.ManagedIndexAction
 import org.opensearch.indexmanagement.indexstatemanagement.transport.action.managedIndex.ManagedIndexRequest
+import org.opensearch.indexmanagement.indexstatemanagement.util.DEFAULT_INDEX_TYPE
 import org.opensearch.indexmanagement.indexstatemanagement.util.FailedIndex
 import org.opensearch.indexmanagement.indexstatemanagement.util.managedIndexConfigIndexRequest
+import org.opensearch.indexmanagement.indexstatemanagement.util.removeClusterStateMetadatas
 import org.opensearch.indexmanagement.opensearchapi.parseFromGetResponse
 import org.opensearch.indexmanagement.settings.IndexManagementSettings
+import org.opensearch.indexmanagement.spi.indexstatemanagement.model.ISMIndexMetadata
 import org.opensearch.indexmanagement.util.IndexUtils
 import org.opensearch.indexmanagement.util.SecurityUtils.Companion.buildUser
 import org.opensearch.indexmanagement.util.SecurityUtils.Companion.userHasPermissionForResource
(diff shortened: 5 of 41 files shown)


src/main/kotlin/org/opensearch/indexmanagement/IndexManagementPlugin.kt
```diff
@@ -230,6 +237,17 @@ class IndexManagementPlugin : JobSchedulerExtension, NetworkPlugin, ActionPlugin
         }
     }
 
+    override fun loadExtensions(loader: ExtensiblePlugin.ExtensionLoader) {
```
thalurur, line 240: we might need to do some validation here - if there is a clash with index type or action parser that have already been installed we should probably skip loading that extension. Can you add a TODO here so we don't miss this?

downsrob: Sure, in this PR I have an error check for index type clashing in the indexMetadataProvider.addMetadataServices method, I can add a similar one in addParsers right now

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/IndexMetadataProvider.kt
```diff
@@ -0,0 +1,42 @@
...
+ */
+class IndexMetadataProvider(
+    val client: Client,
+    val clusterService: ClusterService,
+    val services: MutableMap<String, IndexMetadataService>
+) {
+
+    private val logger = LogManager.getLogger(javaClass)
+
+    suspend fun getIndexMetadata(type: String, indexNames: List<String>): Map<String, ISMIndexMetadata>? {
```
thalurur, line 26: minor: do you think we should set the type defaulted to _default so caller don't need to specify the type unless needed?

Optional: Also do you think it makes sense to resolve the indexName in the class instead of delegating it to the caller?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/IndexMetadataProvider.kt
```diff
@@ -0,0 +1,42 @@
...
+        logger.info("IndexMetadataProvider being called. Providers are ${services.keys}")
```
thalurur, line 27: we can probably remove this?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/IndexMetadataProvider.kt
```diff
@@ -0,0 +1,42 @@
...
+        val service = services[type]
+        if (service != null) {
+            return service.getMetadata(indexNames, client, clusterService)
+        }
+        throw IllegalArgumentException("Index type [$type] was not recognized when trying to get index metadata")
```
thalurur, line 32: optional: I think you can do `getOrThrow()` to simplify this logic

src/main/kotlin/org/opensearch/indexmanagement/IndexManagementPlugin.kt
thalurur: if provider throws an error then we don't want to install parsers as well right and same goes vice versa? or do you think we should partial install

downsrob: No, I agree that we don't want to partially load. I thought that throwing an exception here would make it so the cluster doesn't spin up at all, so partially loading wouldn't be a problem?

thalurur: That is true. I guess that is another option too, to not let the cluster start - which makes more sense instead of silently not picking up the extension cause we found a clash.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/DefaultIndexMetadataService.kt
```diff
@@ -0,0 +1,57 @@
...
+class DefaultIndexMetadataService : IndexMetadataService {
+
+    private val logger = LogManager.getLogger(javaClass)
+
+    /**
+     * Returns the default index metadata needed for ISM
+     */
+    @Suppress("SpreadOperator")
+    override suspend fun getMetadata(indices: List<String>, client: Client, clusterService: ClusterService): Map<String, ISMIndexMetadata> {
+        logger.info("DefaultIndexMetadataService being used to get metadata for $indices")
```
thalurur, line 28: is this for debugging purpose?

downsrob: Yes, that was left from debugging, thanks for calling it out

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/resthandler/RestRetryFailedManagedIndexActionIT.kt
```diff
@@ -207,7 +207,9 @@ class RestRetryFailedManagedIndexActionIT : IndexStateManagementRestTestCase() {
         }
     }
 
-    fun `test index failed`() {
+    // TODO reenable test once allocation action is done
+    @Suppress("UnusedPrivateMember")
+    private fun `test index failed`() {
```
thalurur, line 212: can this test be enabled now?

downsrob: Yes, thanks. When I created the PR the allocation action wasn't done yet

downsrob: There are probably a few others I missed, I will look now, feel free to call them out if you see them though

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -783,4 +784,22 @@ object ManagedIndexRunner :
...
+    /**
+     * Get the index creation date for the first time to cache it on the ManagedIndexMetadata
+     */
+    @Suppress("ReturnCount")
+    private fun getIndexCreationDate(managedIndexConfig: ManagedIndexConfig): Long? {
+        try {
+            val metadata = clusterService.state().metadata()
+            // Check if this index is a hot/warm index in cluster state first
+            if (metadata.hasIndex(managedIndexConfig.index)) {
+                // TODO use MetadataIndexProvider
```
thalurur, line 797: just checking this will be addressed in other PR?

downsrob: Yep, in the runner PR https://github.com/opensearch-project/index-management/blob/feb15a1999c973ce8fca5786b6b59388520c10ff/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt#L830-L842

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
```diff
@@ -387,6 +382,23 @@ class TransportAddPolicyAction @Inject constructor(
...
+                        log.debug("Cleaned cluster state metadata for $indices, ${response.isAcknowledged}")
+                    }
+
+                    override fun onFailure(e: Exception) {
+                        log.error("Failed to clean cluster state metadata for $indices")
+                    }
+                }
+            )
+        }
+
```
thalurur, line 401: do you think we can make this suspend function so you don't have the onResponse and onFailure since we are not doing anything after the call?

downsrob: Added, thanks!

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/updateindexmetadata/TransportUpdateManagedIndexMetaDataAction.kt
```diff
@@ -53,13 +55,27 @@ class TransportUpdateManagedIndexMetaDataAction @Inject constructor(
...
+        val indices = checkExtensionsOverrideBlock(indicesToAddTo + indicesToRemoveFrom, state)
 
         return state.blocks.indicesBlockedException(ClusterBlockLevel.METADATA_WRITE, indices)
     }
 
+    /**
+     * Index Management extensions may provide an index setting, which, if set to true, overrides the cluster metadata write block
+     */
+    private fun checkExtensionsOverrideBlock(indices: Array<Index>, state: ClusterState): Array<String> {
+        val indexBlockOverrideSettings = indexMetadataProvider.getIndexMetadataWriteOverrideSettings()
```
thalurur, line 69: can we instead just hard code this instead of interface method

thalurur (dismissed):

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/AddPolicyRequest.kt
```diff
@@ -10,23 +10,31 @@ import org.opensearch.action.ActionRequestValidationException
...
 
     override fun validate(): ActionRequestValidationException? {
         var validationException: ActionRequestValidationException? = null
         if (indices.isEmpty()) {
             validationException = ValidateActions.addValidationError("Missing indices", validationException)
+        } else if (indexType != DEFAULT_INDEX_TYPE && indices.size > 1) {
+            validationException = ValidateActions.addValidationError(
+                MULTIPLE_INDICES_CUSTOM_INDEX_TYPE_ERROR,
+                validationException
+            )
```
dbbaughe, line 37: Might be susceptible to future developer not knowing they need to add this check to a new API.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
```diff
@@ -131,41 +123,47 @@ class TransportAddPolicyAction @Inject constructor(
...
-                getPolicy()
-            } else {
-                validateAndGetPolicy(0, requestedIndices)
+            getClusterState()
+        }
+
+        @Suppress("SpreadOperator")
+        fun getClusterState() {
+            startTime = Instant.now()
+            CoroutineScope(Dispatchers.IO).launch {
```
bowenlan-amzn, line 132: Curious why do we need coroutine here?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
```diff
@@ -186,13 +184,51 @@ class TransportAddPolicyAction @Inject constructor(
...
+                client.admin()
+                    .cluster()
+                    .state(
+                        clusterStateRequest,
+                        object : ActionListener<ClusterStateResponse> {
+                            override fun onResponse(response: ClusterStateResponse) {
+                                CoroutineScope(Dispatchers.IO).launch { removeClusterStateMetadatas(indicesToAdd.map { Index(it.value, it.key) }) }
+
+                                getUuidsForClosedIndices(response.state).forEach {
+                                    failedIndices.add(FailedIndex(indicesToAdd[it] as String, it, "This index is closed"))
```
bowenlan-amzn, line 220: Do we consider put `whether index is closed` into ISMIndexMetadata

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
```diff
@@ -186,13 +184,51 @@ class TransportAddPolicyAction @Inject constructor(
...
+                    .local(false)
+                    .waitForTimeout(TimeValue.timeValueMillis(ADD_POLICY_TIMEOUT_IN_MILLIS))
+                    .indicesOptions(strictExpandOptions)
+                client.admin()
+                    .cluster()
+                    .state(
+                        clusterStateRequest,
+                        object : ActionListener<ClusterStateResponse> {
+                            override fun onResponse(response: ClusterStateResponse) {
+                                CoroutineScope(Dispatchers.IO).launch { removeClusterStateMetadatas(indicesToAdd.map { Index(it.value, it.key) }) }
```
bowenlan-amzn, line 217: Double confirming here, `removeClusterStateMetadatas` is a best effort, not coupling with any other logic of this class, so we can use coroutine to execute it.

src/main/resources/mappings/opendistro-ism-config.json
```diff
@@ -662,6 +662,10 @@
         "rolled_over": {
           "type": "boolean"
         },
+        "index_creation_date": {
```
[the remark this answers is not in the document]
downsrob: The schema version was already increased in 1.3 with the addition of min_primary_shard_size https://github.com/opensearch-project/index-management/commit/62eada9c89e0cf66c86ae2dcf1d0c76dca9da8f9

build.gradle
```diff
@@ -312,7 +312,8 @@ integTest {
     exclude 'org/opensearch/indexmanagement/indexstatemanagement/action/ActionTimeoutIT.class'
     exclude 'org/opensearch/indexmanagement/indexstatemanagement/action/IndexStateManagementHistoryIT.class'
     exclude 'org/opensearch/indexmanagement/indexstatemanagement/runner/*'
-    exclude 'org/opensearch/indexmanagement/indexstatemanagement/resthandler/*'
+    exclude 'org/opensearch/indexmanagement/indexstatemanagement/resthandler/IndexStateManagementRestApiIT.class'
```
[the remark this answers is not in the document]
downsrob: We don't. Once the rest api and runner PRs are merged, then I will use the diff between this branch and main to check out every new comment to make sure there aren't any missed bits

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/resthandler/RestDeletePolicyActionIT.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.opensearch.indexmanagement.IndexManagementPlugin.Companion.POLICY_BASE_URI
+import org.opensearch.indexmanagement.indexstatemanagement.IndexStateManagementRestTestCase
+import org.opensearch.indexmanagement.makeRequest
+import org.opensearch.indexmanagement.waitFor
+import org.opensearch.rest.RestStatus
+import kotlin.test.assertFailsWith
+
+@Suppress("UNCHECKED_CAST")
+class RestDeletePolicyActionIT : IndexStateManagementRestTestCase() {
+    fun `test delete single policy`() {
```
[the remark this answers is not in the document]
downsrob: You are right. Unfortunately I just saw that there wasn't any integration test named RestDeletePolicyActionIT and wrote a new one when I was refactoring, as I thought that one didn't exist. I'll remove these tests.

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/resthandler/RestGetPolicyActionIT.kt
```diff
@@ -0,0 +1,52 @@
...
+
+import org.opensearch.indexmanagement.IndexManagementPlugin.Companion.POLICY_BASE_URI
+import org.opensearch.indexmanagement.indexstatemanagement.IndexStateManagementRestTestCase
+import org.opensearch.indexmanagement.indexstatemanagement.model.Policy
+import org.opensearch.indexmanagement.makeRequest
+import org.opensearch.indexmanagement.opensearchapi.convertToMap
+import org.opensearch.rest.RestStatus
+
+@Suppress("UNCHECKED_CAST")
+class RestGetPolicyActionIT : IndexStateManagementRestTestCase() {
```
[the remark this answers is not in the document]
downsrob: Yep, I missed that and wrote new ones. Removing.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/IndexMetadataProvider.kt
```diff
@@ -78,6 +78,10 @@ class IndexMetadataProvider(
         metadata
     }
 
+    fun getIndexMetadataWriteOverrideSettings(): List<String> {
```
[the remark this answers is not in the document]
downsrob: Another PR added some comments on this in parallel:
https://github.com/opensearch-project/index-management/blob/development-extension/spi/src/main/kotlin/org.opensearch.indexmanagement.spi/IndexManagementExtension.kt#L51-L52
Here it is used https://github.com/downsrob/index-management/blob/rest-apis/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/updateindexmetadata/TransportUpdateManagedIndexMetaDataAction.kt#L68-L72

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/AddPolicyRequest.kt
downsrob: Fair enough, I can add an additional validation to the IndexMetadataProvider so that check is never missed, but it would be preferable if this exception always occurred during request validation. Let me know if you can think of a better way to enforce this.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
downsrob: I would say no, I think the concept of 'closed' is somewhat specific to in-cluster indices and this interface should extend to non-cluster forms of indices. As an extension to the interface, I believe that we should provide extensions with some validation capabilities, which would encompass this behavior, but that is out of the current scope

build.gradle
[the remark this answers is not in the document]
thalurur: I did sweep before my last PR and didn’t see any related pending TODOs

dbbaughe (approved):

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/DefaultIndexMetadataService.kt
```diff
@@ -38,18 +39,22 @@ class DefaultIndexMetadataService(val customUUIDSetting: String? = null) : Index
 
         response.state.metadata.indices.forEach {
             // TODO waiting to add document count until it is definitely needed
-            val uuid = if (customUUIDSetting != null) {
-                it.value.settings.get(customUUIDSetting, it.value.indexUUID)
-            } else {
-                it.value.indexUUID
-            }
+            val uuid = getCustomIndexUUID(it.value)
```
bowenlan-amzn, line 42: Add comment for `customUUIDSetting` about what it does.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -228,18 +235,35 @@ object ManagedIndexRunner :
...
             return
         }
-        val managedIndexMetaData = indexMetaData.getManagedIndexMetadata(client)
-        val clusterStateMetadata = indexMetaData.getManagedIndexMetadata()
 
-        if (!isMetadataMoved(clusterStateMetadata, managedIndexMetaData, logger)) {
-            logger.info("Skipping execution while pending migration of metadata for ${managedIndexConfig.jobName}")
-            return
+        var indexMetadata = getIndexMetadata(managedIndexConfig.index)
+        // The index uuids not matching can happen when two jobs share a name but have different index types
```
bowenlan-amzn, line 245: Refactor this part, we are doing 2 things here:
1. if the index uuid from cluster state doesn't match the index uuid in managedIndexConfig, we set the indexMetadata to null
2. we do a sanity check: try to find this index from all the other extension provider, if cannot find, we return early.

downsrob: I added some additional comments for this but I am going to delay further refactoring, I think that with the comments it helps clarify the two steps

dbbaughe (dismissed):

bowenlan-amzn (approved):

bowenlan-amzn (approved):
👏

dbbaughe (approved):

Issue #249, opened 2022-01-14
[Distribution Build Failure] Detekt fails to resolve com.beust:jcommander:1.74


downsrob:
**Describe the bug**
Detekt 1.0.0-RC15, used in the index management build process for detecting code smells, fails to run due to not being able to resolve the detekt dependency jcommander 1.74. This dependency, jcommander 1.74, does not exist on maven central, and can only be resolved through jcenter. As jcenter is sunset, this dependency may never be resolved, so detekt must be upgraded to a version which doesn't include this dependency, or disabled. In later versions of detekt, the jcommander dependency was moved to version 1.72, so upgrading detekt should resolve this issue.

As upgrading detekt adds new code style checks, additional work will be needed to fix the newly found code smells to enable the detekt check to pass.

**To Reproduce**
./gradlew build or ./gradlew detekt fail due to a missing detekt dependency
```
> Task :detekt FAILED
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':detekt'.
> Could not resolve all files for configuration ':detekt'.
   > Could not find com.beust:jcommander:1.74.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/com/beust/jcommander/1.74/jcommander-1.74.pom
       - file:/home/runner/.m2/repository/com/beust/jcommander/1.74/jcommander-1.74.pom
       - https://aws.oss.sonatype.org/content/repositories/snapshots/com/beust/jcommander/1.74/jcommander-1.74.pom
     Required by:
         project : > io.gitlab.arturbosch.detekt:detekt-cli:1.0.0-RC15
```

**Remediation**
To fix this issue:

- [x] Upgrade detekt to a version which uses jcommander 1.72
- [x] Backport the upgraded detekt version to the 1.2 branch, raise the allowed weighted issues count to enable building OR disable detekt entirely on the 1.2 branch https://github.com/opensearch-project/index-management/pull/250
- [x] Fix the code style issues which are newly recognized in the upgraded detekt version on main https://github.com/opensearch-project/index-management/pull/252

**Additional context**
As jcommander 1.74 is not on maven central, the jcenter workaround described here https://github.com/opensearch-project/opensearch-build/issues/1456 does not help resolve the dependency.

PR #252, opened 2022-01-18
Upgrades detekt version to 1.17.1

downsrob:

*Issue #, if available:*
https://github.com/opensearch-project/index-management/issues/249

*Description of changes:*

- Upgrades detekt version from 1.0.0-RC15 to 1.17.1
- Adds exclusion for some rules in test files and for swallowed exceptions
- Fixes var could be val issues
- Removes unused variables
- Moves comments inside of empty function bodies to satisfy detekt rule
- Changes for loops with unused iterators to repeat() calls with stubbed variables to satisfy unused local variable rule

*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/IndexManagementIndices.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/IndexManagementIndices.kt
@@ -78,7 +78,6 @@ class IndexManagementIndices(
     /**
      * Attempt to create [INDEX_MANAGEMENT_INDEX] and return whether it exists
      */
-    @Suppress("SwallowedException")
     suspend fun attemptInitStateManagementIndex(client: Client): Boolean {
         if (indexManagementIndexExists()) return true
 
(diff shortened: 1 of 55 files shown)


bowenlan-amzn (approved):

src/main/kotlin/org/opensearch/indexmanagement/IndexManagementIndices.kt
```diff
@@ -78,6 +78,7 @@ class IndexManagementIndices(
     /**
      * Attempt to create [INDEX_MANAGEMENT_INDEX] and return whether it exists
      */
+    @Suppress("SwallowedException")
```
thalurur, line 81: just curious - do we need to add this? since it seems the `.yml` sets is `active: false`

thalurur (dismissed):

src/main/kotlin/org/opensearch/indexmanagement/IndexManagementIndices.kt
downsrob: Good catch, I think I suppressed this one and then on the next one I looked at I realized that it didn't work for us because detekt just doesn't pick up our logging in the catch blocks. I'll remove this.

thalurur (approved):

bowenlan-amzn (approved):
Thanks!

PR #255, opened 2022-01-19
Not managing indices when matched certain pattern

thalurur:

*Description of changes:*
Not managing indices that match certain blocked patterns

*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
@@ -282,6 +282,10 @@ class TransportAddPolicyAction @Inject constructor(
                 failedIndices.add(FailedIndex(indicesToAdd[it] as String, it, "This index is closed"))
                 indicesToAdd.remove(it)
             }
+
+            // Removing all the unmanageable Indices
+            indicesToAdd.entries.removeIf { (_, indexName) -> IndexUtils.isUnManageableIndexPattern(indexName) }
+
             if (indicesToAdd.isEmpty()) {
                 actionListener.onResponse(ISMStatusResponse(0, failedIndices))
                 return
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
@@ -58,6 +58,7 @@ import org.opensearch.indexmanagement.indexstatemanagement.transport.action.mana
 import org.opensearch.indexmanagement.indexstatemanagement.transport.action.managedIndex.ManagedIndexRequest
 import org.opensearch.indexmanagement.indexstatemanagement.util.FailedIndex
 import org.opensearch.indexmanagement.indexstatemanagement.util.IndexEvaluator
+import org.opensearch.indexmanagement.indexstatemanagement.util.IndexEvaluator.Companion.EVALUATION_FAILURE_MESSAGE
 import org.opensearch.indexmanagement.indexstatemanagement.util.managedIndexConfigIndexRequest
 import org.opensearch.indexmanagement.opensearchapi.parseFromGetResponse
 import org.opensearch.indexmanagement.settings.IndexManagementSettings
@@ -289,7 +290,7 @@ class TransportAddPolicyAction @Inject constructor(
             indicesToAdd.entries.removeIf { (uuid, indexName) ->
                 val shouldRemove = indexEvaluator.isUnManageableIndex(indexName)
                 if (shouldRemove) {
-                    failedIndices.add(FailedIndex(indexName, uuid, "Matches restricted index pattern defined in the cluster setting"))
+                    failedIndices.add(FailedIndex(indexName, uuid, EVALUATION_FAILURE_MESSAGE))
                 }
                 shouldRemove
             }
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
@@ -57,6 +57,7 @@ import org.opensearch.indexmanagement.indexstatemanagement.transport.action.ISMS
 import org.opensearch.indexmanagement.indexstatemanagement.transport.action.managedIndex.ManagedIndexAction
 import org.opensearch.indexmanagement.indexstatemanagement.transport.action.managedIndex.ManagedIndexRequest
 import org.opensearch.indexmanagement.indexstatemanagement.util.FailedIndex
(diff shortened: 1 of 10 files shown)


src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/resthandler/RestAddPolicyActionIT.kt
```diff
@@ -137,11 +137,23 @@ class RestAddPolicyActionIT : IndexStateManagementRestTestCase() {
...
         createIndex(indexThree, null)
+        if (!indexExists(indexFour)) {
+            createIndex(indexFour)
+        }
+        if (!indexExists(indexFive)) {
+            createIndex(indexFive)
+        }
+        if (!indexExists(indexSix)) {
+            createIndex(indexSix)
+        }
```
aditjind, line 156: Should we replace this list of `if's`?

src/test/kotlin/org/opensearch/indexmanagement/util/IndexUtilsTests.kt
```diff
@@ -88,4 +89,16 @@ class IndexUtilsTests : OpenSearchTestCase() {
...
         assertFalse(shouldUpdateIndex)
     }
+
+    fun `test security index and kibana should not be manageable`() {
+        assertTrue(IndexUtils.isUnManageableIndexPattern(".opendistro-security-config"))
+        assertTrue(IndexUtils.isUnManageableIndexPattern(".kibana_1"))
+        assertTrue(IndexUtils.isUnManageableIndexPattern(".kibana"))
+        assertTrue(IndexUtils.isUnManageableIndexPattern(".kibana_20"))
+        assertTrue(IndexUtils.isUnManageableIndexPattern(".kibana_022"))
+        assertFalse(IndexUtils.isUnManageableIndexPattern(".kibana_i2"))
```
aditjind, line 99: Error message in the assertion?

thalurur: added

src/main/kotlin/org/opensearch/indexmanagement/util/IndexUtils.kt
```diff
@@ -191,5 +193,13 @@ class IndexUtils {
             val byteArray = ByteBuffer.allocate(BYTE_ARRAY_SIZE).putLong(hash.h1).putLong(hash.h2).array()
             return Base64.getUrlEncoder().withoutPadding().encodeToString(byteArray)
         }
+
+        /**
+         * Check if the index is unmanageable by policy
+         */
+        fun isUnManageableIndexPattern(index: String): Boolean {
+            // TODO: We can potentially expose this as setting if required
+            return Regex("\\.opendistro_security|\\.kibana.*|\\$INDEX_MANAGEMENT_INDEX").matches(index)
```
downsrob, line 202: I am curious as to why this works? I would think that it would be \.opendistro_security rather than \\.opendistro_security. Is it escaping the escape character or something?

downsrob (approved):

src/main/kotlin/org/opensearch/indexmanagement/util/IndexUtils.kt
thalurur: `\` is a special character in java and interpreted as escape sequence - https://stackoverflow.com/questions/35237697/error-illegal-escape-character-when-trying-to-create-a-path

downsrob: Thanks. As we are exposing this as a cluster setting do you think that we should sanitize the setting string in some way to allow a user to input a normal regex? Or do you think it is not a big concern because all it would need is a note in the docs saying to escape backslashes?

thalurur: I tested setting value using single `\.` and its not considered  a valid string and throw error since it seems to be parsed to java string before storing

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
```diff
@@ -282,6 +284,16 @@ class TransportAddPolicyAction @Inject constructor(
                 failedIndices.add(FailedIndex(indicesToAdd[it] as String, it, "This index is closed"))
                 indicesToAdd.remove(it)
             }
+
+            // Removing all the unmanageable Indices
+            indicesToAdd.entries.removeIf { (uuid, indexName) ->
+                val shouldRemove = indexEvaluator.isUnManageableIndex(indexName)
+                if (shouldRemove) {
+                    failedIndices.add(FailedIndex(indexName, uuid, "Matches restricted index pattern"))
```
bowenlan-amzn, line 292: Can we mention `restricted_index_pattern` cluster setting here, otherwise it maybe confusing what this means.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
```diff
@@ -282,6 +284,16 @@ class TransportAddPolicyAction @Inject constructor(
                 failedIndices.add(FailedIndex(indicesToAdd[it] as String, it, "This index is closed"))
                 indicesToAdd.remove(it)
             }
+
+            // Removing all the unmanageable Indices
+            indicesToAdd.entries.removeIf { (uuid, indexName) ->
+                val shouldRemove = indexEvaluator.isUnManageableIndex(indexName)
+                if (shouldRemove) {
+                    failedIndices.add(FailedIndex(indexName, uuid, "Matches restricted index pattern defined in the cluster setting"))
```
downsrob, line 292: Could you make this a constant on either the TransportAddPolicyAction or the IndexEvaluator class?

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/resthandler/RestAddPolicyActionIT.kt
```diff
@@ -176,4 +179,60 @@ class RestAddPolicyActionIT : IndexStateManagementRestTestCase() {
...
+        val actualMessage = response.asMap()
+        // Not going to attach policy to ism config index or other restricted index patterns
+        val expectedMessage = mapOf(
+            UPDATED_INDICES to 1,
+            FAILURES to true,
+            FAILED_INDICES to listOf(
+                mapOf(
+                    "index_name" to indexOne,
+                    "index_uuid" to getUuid(indexOne),
+                    "reason" to "Matches restricted index pattern defined in the cluster setting"
```
downsrob, line 211: Could you use the EVALUATION_FAILURE_MESSAGE constant in these tests as well?

thalurur: No I wanted to keep this as string explicitly since this would be a response in API, wanted some test failure in case we change the message in future

downsrob: Fair point, I can definitely see such a situation occurring. Thanks

downsrob (approved):

bowenlan-amzn (approved):

PR #262, opened 2022-02-07
Refactors the managed index runner to work with extensions

downsrob:
*Issue #, if available:*
NA

*Description of changes:*
- Adds NewClusterEventListeners so that extensions may react to cluster events with the NewCluster type
- Supports transition condition evaluation for non-default index types by using the IndexMetadataProvider to get the index creation date. Adds caching of the index creation date for all index types to the ManagedIndexRunner
- Adds a setting to deleteIndexMetadataAfterFinish to actions. When set to true, the managed index config and metadata for the index that the policy/action is attached to will be deleted following action completion. For in-cluster indices this should be handled automatically, so this setting is just to support this behavior for custom actions on non-default index types.
- Adds IndexMetadataProvider implementation from https://github.com/opensearch-project/index-management/pull/245 Changes to the IndexMetadataProvider may need to be made to either PR based on the other's comments. This is awkward, but it is a blocker otherwise.
- Code cleanup

*CheckList:*


src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/runner/ManagedIndexRunnerIT.kt
```diff
@@ -141,11 +168,18 @@ class ManagedIndexRunnerIT : IndexStateManagementRestTestCase() {
...
+            // block the read_only action
+            val blockedActions = "[\"${ReadOnlyAction.name}\"]"
+            updateClusterSetting(ManagedIndexSettings.BLOCKED_ACTIONS_LIST.key, blockedActions, escapeValue = false)
+        } else {
+            // remove read_only from the allowlist
+            val allowedActions = ISMActionsParser.instance.parsers.map { it.getActionType() }.toList()
+                .filter { actionType -> actionType != ReadOnlyAction.name }
+                .joinToString(prefix = "[", postfix = "]") { string -> "\"$string\"" }
+            updateClusterSetting(ManagedIndexSettings.ALLOW_LIST.key, allowedActions, escapeValue = false)
+        }
```
thalurur, line 182: why do we need both of these here?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/settings/ManagedIndexSettings.kt
```diff
@@ -175,6 +176,14 @@ class ManagedIndexSettings {
             Setting.Property.Dynamic
```
thalurur, line 176: Can we mark this as deprecated setting and to use blocked_actions_list instead?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/Policy.kt
```diff
@@ -116,14 +116,15 @@ data class Policy(
...
      */
-    fun getDisallowedActions(allowList: List<String>): List<String> {
-        val allowListSet = allowList.toSet()
+    fun getDisallowedActions(allowList: List<String>, blockedActionsList: List<String>): List<String> {
+        val allowedActionsSet = allowList.toSet() - blockedActionsList.toSet()
         val disallowedActions = mutableListOf<String>()
         this.states.forEach { state ->
             state.actions.forEach { actionConfig ->
-                if (!allowListSet.contains(actionConfig.type)) {
+                if (!allowedActionsSet.contains(actionConfig.type)) {
```
thalurur, line 127: can we not just do `!blockedActionsList.conatins(actionConfig.type)`

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/Policy.kt
```diff
@@ -116,14 +116,15 @@ data class Policy(
...
+     * Disallowed actions are specified in the [ManagedIndexSettings.BLOCKED_ACTIONS_LIST], or are not specified in the
+     * [ManagedIndexSettings.ALLOW_LIST] setting.
      */
-    fun getDisallowedActions(allowList: List<String>): List<String> {
-        val allowListSet = allowList.toSet()
+    fun getDisallowedActions(allowList: List<String>, blockedActionsList: List<String>): List<String> {
+        val allowedActionsSet = allowList.toSet() - blockedActionsList.toSet()
         val disallowedActions = mutableListOf<String>()
         this.states.forEach { state ->
             state.actions.forEach { actionConfig ->
```
thalurur, line 126: nit: `actionConfig -> action`

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/settings/ManagedIndexSettings.kt
downsrob: Added the deprecated settings property, thanks

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/Policy.kt
downsrob: Yes, so we can continue to support the allow list setting while giving the deprecated warning until it is removed in the next major version

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/runner/ManagedIndexRunnerIT.kt
downsrob: Though the allow list setting is getting deprecated, I wanted to keep a test case for it to make sure that we maintain support until it is removed.

downsrob: I actually had to remove this, as I couldn't find a way to successfully suppress the deprecation warning when updating the setting

spi/src/main/kotlin/org.opensearch.indexmanagement.spi/indexstatemanagement/Action.kt
```diff
@@ -72,6 +90,25 @@ abstract class Action(
...
+     */
+    final fun isFinishedSuccessfully(managedIndexMetaData: ManagedIndexMetaData): Boolean {
+        val policyRetryInfo = managedIndexMetaData.policyRetryInfo
+        if (policyRetryInfo == null || policyRetryInfo.failed) return false
+        val actionMetaData = managedIndexMetaData.actionMetaData
+        if (actionMetaData == null || actionMetaData.failed || actionMetaData.name != this.type) return false
+        val stepMetaData = managedIndexMetaData.stepMetaData
+        if (stepMetaData == null || !isLastStep(stepMetaData.name) || stepMetaData.stepStatus != Step.StepStatus.COMPLETED) return false
+        return true
+    }
```
dbbaughe, line 104: Is this new code or just existing code refactored to this place?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -383,6 +414,14 @@ object ManagedIndexRunner :
                 return
             }
 
+            if (action.isFinishedSuccessfully(executedManagedIndexMetaData)) {
+                // Custom actions added by extensions may require a manual deletion of the managed index metadata after a successful execution
```
dbbaughe, line 418: Could you clarify when this happens?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -497,7 +536,7 @@ object ManagedIndexRunner :
         }
     }
 
-    private fun getFailedInitializedManagedIndexMetaData(
+    private suspend fun getFailedInitializedManagedIndexMetaData(
```
dbbaughe, line 539: Why's it a suspend fun now?

spi/src/main/kotlin/org.opensearch.indexmanagement.spi/indexstatemanagement/Action.kt
downsrob: It's a generalization of the isSuccessfulDelete function:
https://github.com/opensearch-project/index-management/blob/main/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/util/ManagedIndexUtils.kt#L430-L433
The purpose is that if an extension creates an action to delete their off cluster index, then they can set deleteIndexMetadataAfterFinish to true in the custom action, and then after it isFinishedSuccessfully, the metadata will be deleted for the action. 
Thinking about this a bit more now, I can add an additional validation [here](https://github.com/opensearch-project/index-management/blob/main/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/State.kt#L30-L33) to make it so these extension actions which set deleteIndexMetadataAfterFinish = true can't transition or have actions after the delete

downsrob: I do see a logical error in the code though, `if (policyRetryInfo == null || policyRetryInfo.failed) return false` would mean that the action would never be noted as isFinishedSuccessfully if the policyRetryInfo was null. It is initialized with the metadata so it shouldn't be a problem but I will change that as well

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
downsrob: It is suspend as it calls getIndexCreationDate, which was made into a suspend function because of using `indexMetadataProvider.getMultiTypeISMIndexMetadata` to get the indexCreationDate

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
downsrob: I am removing the word manual from the comment as that is misleading. I added a bit on this in the other comment, but if an extension creates a deletion action for an off cluster index, we will not pick up the index deletion in the coordinator like we do for our delete action. Instead, this is new logic to allow actions to designate that the index metadata should be deleted after the action is finished. I will improve the comments around this

dbbaughe (dismissed):

dbbaughe (approved):

bowenlan-amzn (approved):
Good good

PR #267, opened 2022-02-16
Updates search text field to keyword subfield for policies and managed indices

dbbaughe:
*Issue #, if available:*
https://github.com/opensearch-project/index-management-dashboards-plugin/issues/117
https://github.com/opensearch-project/index-management-dashboards-plugin/issues/119
https://github.com/opensearch-project/index-management-dashboards-plugin/issues/112

*Description of changes:*
The mappings for the managed index and policy name/id fields were going through the standard analyzer which meant an index or policy with a name like "foo-bar-baz" would be broken up into tokens: ["foo", "bar", "baz"] and there would be no more hyphen. That combined with our use of the query string query which would also incorrectly use a hyphen as a negation (sometimes) and other unwanted behavior with wildcards being prepended/appended would cause searches on the UI to return empty results. e.g. searching "foo-*" would not return "foo-bar-baz".

So instead we'll query the keyword which should have stayed the same, added test cases to make sure hyphens and periods at least are working as we expect them to, and all other tests were passing so hoping no unknown regressions..

Tested on the UI with no changes and searches are working as expected now when the search bar includes hyphens or periods for datastreams.

*CheckList:*


src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/resthandler/RestExplainActionIT.kt
```diff
@@ -126,6 +126,114 @@ class RestExplainActionIT : IndexStateManagementRestTestCase() {
...
+        waitFor {
+            val expected = mapOf(
+                indexName1Map,
+                indexName2Map,
+                indexName4Map,
+                indexName5Map,
+                "total_managed_indices" to 4
+            )
+            // These should match all non datastream managed indices
+            assertResponseMap(expected, getExplainMap(indexName = null, queryParams = "queryString=$testIndexName*"))
```
thalurur, line 195: nit: hardening test a little bit, a case with two space separated tokens so we are covering AND condition we are enforcing always

thalurur (approved):

dbbaughe:
Discussed with Ravi offline, if someone was taking advantage of the negation part of the query string, such as return me all managed indices that do not have the token -foobar in the name, then this would break that as the new behavior would be return me all managed indices that do not have an index name of foobar (as it's exact now). That being said... I don't think this is a behavior we ever went out to support, so it's more akin to a bug becoming a feature if someone did use it this way. The original intent was always just a simple query string that lets you search up index names and use a glob * pattern to match any number of characters. We plan to introduce this currently in the 2.0 version.

praveensameneni (dismissed):

downsrob (approved):

thalurur (approved):

PR #272, opened 2022-02-23
Refactor coordinator to support create and delete index events when different index types exist on cluster

thalurur:
*Issue #, if available:*
N/A

*Description of changes:*
Adding support for coordinator to correctly determine the indices that should be managed/delete currently managed indices by checking all of the metadata services to get the existing indices uuids and there existence in the cluster - rather than relying on cluster state metadata alone

*CheckList:*


spi/src/main/kotlin/org.opensearch.indexmanagement.spi/indexstatemanagement/IndexMetadataService.kt
```diff
@@ -14,5 +14,10 @@ interface IndexMetadataService {
     /**
```
bowenlan-amzn, line 14: For this interface, can we have some doc at top, mentioning each implementation is for one unique IndexType. Since IndexType is sth. we invent here, also worth some explanation for understanding.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/IndexMetadataProvider.kt
```diff
@@ -0,0 +1,67 @@
...
+        clusterService.clusterSettings.addSettingsUpdateConsumer(ManagedIndexSettings.RESTRICTED_INDEX_PATTERN) {
+            restrictedIndexPattern = it
+        }
+    }
+
+    fun isUnManageableIndex(index: String): Boolean {
+        return Regex(restrictedIndexPattern).matches(index)
+    }
+
+    suspend fun getISMIndexMetadata(type: String = DEFAULT_INDEX_TYPE, indexNames: List<String>): Map<String, ISMIndexMetadata> {
```
bowenlan-amzn, line 39: Suggest using `getISMIndexMetadataForIndexType`, otherwise it could be confusing to compare with the below `getAllISMIndexMetadata` which aggregate all indextypes

spi/src/main/kotlin/org.opensearch.indexmanagement.spi/indexstatemanagement/IndexMetadataService.kt
thalurur: Sure let me add it

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
```diff
@@ -294,7 +295,9 @@ class ManagedIndexCoordinator(
         var indicesToClean = emptyList<Index>()
         if (event.indicesDeleted().isNotEmpty()) {
             val managedIndices = getManagedIndices(event.indicesDeleted().map { it.uuid })
-            indicesToClean = event.indicesDeleted().filter { it.uuid in managedIndices.keys }
+            val allIndicesUuid = indexMetadataProvider.getAllISMIndexMetadata().map { it.indexUuid }
```
downsrob, line 298: This could mean a large get all request to all metadata providers, which adds overhead per indices deleted event. Can we just search the index name for each provider, or do we need to get all indices each time?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
```diff
@@ -582,33 +588,36 @@ class ManagedIndexCoordinator(
...
 
         lastFullSweepTimeNano = System.nanoTime()
     }
 
-    private fun getAllClusterIndices(): List<Index> {
-        return clusterService.state().metadata.indices.values().map { it.value.index }.distinct().filterNotNull()
-    }
-
-    private fun getUnManagedIndices(allIndices: List<Index>, managedIndexUuids: Set<String>): List<Index> {
-        return allIndices.filter { it.uuid !in managedIndexUuids && !indexEvaluator.isUnManageableIndex(it.name) }
```
downsrob, line 611: We are no longer checking if the index is unmanageable here, is this a logical change or did you decide that it wasn't necessary?

downsrob:
I noticed the failure, `DeleteActionIT.test basic` 
```
java.lang.NullPointerException: null cannot be cast to non-null type kotlin.collections.Map<*, *>
...
 at org.opensearch.indexmanagement.indexstatemanagement.action.DeleteActionIT$test basic$4.invoke(DeleteActionIT.kt:64)
```
I had this flaky failure on another one of my development-extension PRs, we may have introduced some new flakiness in this test with our changes. I ran this IT(only this single test) 100 times yesterday and couldn't replicate the failure.

src/main/kotlin/org/opensearch/indexmanagement/IndexManagementPlugin.kt
```diff
@@ -322,6 +340,15 @@ class IndexManagementPlugin : JobSchedulerExtension, NetworkPlugin, ActionPlugin
                 indexManagementIndices
             )
 
+        indexMetadataProvider = IndexMetadataProvider(
+            settings, client, clusterService,
+            hashMapOf(
+                DEFAULT_INDEX_TYPE to DefaultIndexMetadataService
+                ()
```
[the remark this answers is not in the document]
thalurur: done

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/IndexMetadataProvider.kt
```diff
@@ -0,0 +1,67 @@
...
+
+    fun addMetadataServices(newServices: Map<String, IndexMetadataService>) {
+        val duplicateIndexType = newServices.keys.firstOrNull { services.containsKey(it) }
+        if (duplicateIndexType != null) {
+            throw IllegalArgumentException(getDuplicateServicesMessage(duplicateIndexType))
+        }
+        services.putAll(newServices)
+    }
+
+    suspend fun getAllISMIndexMetadata(): Set<ISMIndexMetadata> {
```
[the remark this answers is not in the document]
thalurur: hmm.. I think the difference is I getting all the indices right? this method still requires the caller to specify the indices to get?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
thalurur: That is fair - I can limit the call to just get the deleted indices don't need to get every index accessible to cluster here

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
thalurur: it was unnecessary since we do this check at the time of policy matching which happens in `createManagedIndexRequests` the step we do immediately after finding all unmanaged indices line 595

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/IndexMetadataProvider.kt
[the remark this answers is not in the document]
thalurur: makes sense to parallelize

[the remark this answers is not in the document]
downsrob: Yeah, I agree, and I see the benefit of having such a function. I only meant that you may want to use similar asynchronous logic to avoid this taking a longer time for users with many index types.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt
downsrob: Great, when I looked through the code I thought the same thing but wanted to call it out as it is easy to miss such things in refactoring

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/DefaultIndexMetadataService.kt
```diff
@@ -0,0 +1,57 @@
...
+            .indices(*indices.toTypedArray())
+            .metadata(true)
+            .local(false)
+            .waitForTimeout(TimeValue.timeValueMillis(DEFAULT_GET_METADATA_TIMEOUT_IN_MILLIS))
+            .indicesOptions(strictExpandOptions)
+
+        val response: ClusterStateResponse = client.suspendUntil { client.admin().cluster().state(clusterStateRequest, it) }
+
+        response.state.metadata.indices.forEach {
+            // TODO waiting to add document count until it is definitely needed
```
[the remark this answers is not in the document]
thalurur: Lets check after all code is merged if we want to remove it or keep it - one concern is with evaluating conditions in transition which has the document count, I know we are skipping this evaluation at the moment in transition step

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/IndexMetadataProvider.kt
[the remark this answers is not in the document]
thalurur: I also copied this method to the PR to help with the delete event processing

thalurur:
> I noticed the failure, `DeleteActionIT.test basic`
> 
> ```
> java.lang.NullPointerException: null cannot be cast to non-null type kotlin.collections.Map<*, *>
> ...
>  at org.opensearch.indexmanagement.indexstatemanagement.action.DeleteActionIT$test basic$4.invoke(DeleteActionIT.kt:64)
> ```
> 
> I had this flaky failure on another one of my development-extension PRs, we may have introduced some new flakiness in this test with our changes. I ran this IT(only this single test) 100 times yesterday and couldn't replicate the failure.

I see, let me check if there is something we are doing incorrect

downsrob (approved):

bowenlan-amzn (approved):

PR #273, opened 2022-02-23
Adds custom action parsing

downsrob:
*Issue #, if available:*
N/A

*Description of changes:*
- Adds a wrapper "custom" field to the ism config to enable custom actions to be stored without needing to update the mappings
- Removes the custom wrapper from the responses of the create/update policy response, and the get policy/policies responses
- Adds parser unit tests
- Adds a customAction boolean to the abstract action, this boolean is set at parse time

*CheckList:*

### DIFF
--- a/spi/src/main/kotlin/org.opensearch.indexmanagement.spi/indexstatemanagement/Action.kt
+++ b/spi/src/main/kotlin/org.opensearch.indexmanagement.spi/indexstatemanagement/Action.kt
@@ -16,7 +16,8 @@ import org.opensearch.indexmanagement.spi.indexstatemanagement.model.StepContext
 
 abstract class Action(
     val type: String,
-    val actionIndex: Int
+    val actionIndex: Int,
+    val customAction: Boolean = false
 ) : ToXContentObject, Writeable {
 
     var configTimeout: ActionTimeout? = null
@@ -26,8 +27,9 @@ abstract class Action(
         builder.startObject()
         configTimeout?.toXContent(builder, params)
         configRetry?.toXContent(builder, params)
-        // TODO: We should add "custom" object wrapper based on the params
+        if (customAction && !params.paramAsBoolean(EXCLUDE_CUSTOM_FIELD_PARAM, false)) builder.startObject(CUSTOM_ACTION_FIELD)
         populateAction(builder, params)
+        if (customAction && !params.paramAsBoolean(EXCLUDE_CUSTOM_FIELD_PARAM, false)) builder.endObject()
         return builder.endObject()
     }
 
@@ -70,5 +72,7 @@ abstract class Action(
 
     companion object {
         const val DEFAULT_RETRIES = 3L
+        const val CUSTOM_ACTION_FIELD = "custom"
+        const val EXCLUDE_CUSTOM_FIELD_PARAM = "exclude_custom"
     }
 }
--- a/spi/src/main/kotlin/org.opensearch.indexmanagement.spi/indexstatemanagement/Action.kt
+++ b/spi/src/main/kotlin/org.opensearch.indexmanagement.spi/indexstatemanagement/Action.kt
@@ -16,12 +16,12 @@ import org.opensearch.indexmanagement.spi.indexstatemanagement.model.StepContext
 
 abstract class Action(
     val type: String,
-    val actionIndex: Int,
-    val customAction: Boolean = false
+    val actionIndex: Int
 ) : ToXContentObject, Writeable {
 
     var configTimeout: ActionTimeout? = null
     var configRetry: ActionRetry? = ActionRetry(DEFAULT_RETRIES)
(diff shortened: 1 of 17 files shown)


thalurur (dismissed):

src/main/resources/mappings/opendistro-ism-config.json
```diff
@@ -429,6 +429,10 @@
                       }
                     }
                   }
+                },
+                "custom": {
+                  "enabled": false,
```
bowenlan-amzn, line 434: https://www.elastic.co/guide/en/elasticsearch/reference/current/enabled.html

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ISMActionsParser.kt
```diff
@@ -77,10 +79,25 @@ class ISMActionsParser private constructor() {
             when (type) {
                 ActionTimeout.TIMEOUT_FIELD -> timeout = ActionTimeout.parse(xcp)
                 ActionRetry.RETRY_FIELD -> retry = ActionRetry.parse(xcp)
+                Action.CUSTOM_ACTION_FIELD -> {
```
bowenlan-amzn, line 82: use `ensureExpectedToken` to explicitly indicate what Token we want next.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ISMActionsParser.kt
```diff
@@ -77,10 +79,25 @@ class ISMActionsParser private constructor() {
...
+                    xcp.nextToken()
+                    val customActionType = xcp.currentName()
+                    xcp.nextToken()
+                    val customParser = parsers.firstOrNull { it.getActionType() == customActionType }
+                    if (customParser != null) {
+                        action = customParser.fromXContent(xcp, totalActions)
+                        action.customAction = customParser.customAction
```
bowenlan-amzn, line 89: For the customAction attribute of parser, are we missing to set it when [addParser](https://github.com/opensearch-project/index-management/pull/273/files#diff-95a7ae870da97f952e874adbb69a690aaa23fa05a40929145bfbe4c814ad50a3R55)?

spi/src/main/kotlin/org.opensearch.indexmanagement.spi/indexstatemanagement/Action.kt
```diff
@@ -21,13 +21,15 @@ abstract class Action(
...
     var configTimeout: ActionTimeout? = null
     var configRetry: ActionRetry? = ActionRetry(DEFAULT_RETRIES)
+    var customAction: Boolean = false
 
     final override fun toXContent(builder: XContentBuilder, params: ToXContent.Params): XContentBuilder {
         builder.startObject()
         configTimeout?.toXContent(builder, params)
         configRetry?.toXContent(builder, params)
-        // TODO: We should add "custom" object wrapper based on the params
+        if (customAction && !params.paramAsBoolean(EXCLUDE_CUSTOM_FIELD_PARAM, false)) builder.startObject(CUSTOM_ACTION_FIELD)
```
bowenlan-amzn, line 30: Add comment about when we want to exclude, when we want to include custom.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ISMActionsParser.kt
downsrob: This will be added in another PR. The logic to get and add extension's parsers is not included in this PR as we don't want to have it in so many parallel reviews. It will just require adding the boolean to the next PR and we will have another review after when this is merged to main

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ISMActionsParser.kt
downsrob: Done, thanks. I did some additional refactoring as well, because I do agree that it was way too difficult to read.

src/main/resources/mappings/opendistro-ism-config.json
downsrob: Keeping it false. This just means that the objects added inside of this custom wrapper won't be indexed in opensearch, so we won't be able to directly search on it, you have to parse the source to get it. Based on our current use cases, indexing it is not required.

spi/src/main/kotlin/org.opensearch.indexmanagement.spi/indexstatemanagement/Action.kt
downsrob: Done!

thalurur (approved):

bowenlan-amzn (approved):

PR #275, opened 2022-02-24
Porting additional missing logic

downsrob:
*Issue #, if available:*
N/A
*Description of changes:*
- Adds index creation date to ManagedIndexMetaData
- Refactors metadata check and adds check for corrupt metadata
- Provides some refactoring of explain, retry failed, and change policy
All of these changes would be included when merging the ism decoupling code. Preemptively merging this will make for an easier review later.

*CheckList:*


thalurur (approved):

src/main/resources/mappings/opendistro-ism-config.json
```diff
@@ -658,6 +658,10 @@
         "rolled_over": {
           "type": "boolean"
         },
+        "index_creation_date": {
```
bowenlan-amzn, line 661: Theoretically we need to bump the version of mapping if making any change in the mapping. @dbbaughe correct me if wrong.

downsrob: The schema version was already increased in 1.3 with the addition of min_primary_shard_size, so we shouldn't have to increase it again if this is going out in 1.3 https://github.com/opensearch-project/index-management/commit/62eada9c89e0cf66c86ae2dcf1d0c76dca9da8f9

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/transition/AttemptTransitionStep.kt
```diff
@@ -136,6 +136,21 @@ class AttemptTransitionStep(
...
     }
 
+    @Suppress("ReturnCount")
+    private fun getIndexCreationDate(): Long {
+        try {
+            // If we do have an index creation date cached already then use that
+            if (indexCreationDate != null) return indexCreationDate
+
+            val clusterStateMetadata = clusterService.state().metadata()
+            clusterStateMetadata.index(indexName).creationDate
```
dbbaughe, line 146: Missing a return statement here?
Might need to handle `clusterStateMetadata.index(indexName)` returning null (or does it have a different behavior)?
And do we need a try catch around this block of code?

downsrob: Good catch, I extracted this from the decoupled code to make the final PR easier and missed this. The try/catch is leftover from that. 
https://github.com/opensearch-project/index-management/blob/feb15a1999c973ce8fca5786b6b59388520c10ff/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/transition/AttemptTransitionStep.kt#L148

downsrob: I think that code also needs to handle `clusterStateMetadata.index(indexName)` returning null so I will edit that in the Runner PR as well.

downsrob:
Saw a couple of flaky failures in the multi node test workflow
```
REPRODUCE WITH: ./gradlew ':integTest' --tests "org.opensearch.indexmanagement.transform.TransformRunnerIT.test no-op execution when no buckets have been modified" -Dtests.seed=70B46EE4D84B4F3E -Dtests.security.manager=false -Dtests.locale=bg -Dtests.timezone=America/Godthab -Druntime.java=14
  2> java.lang.NullPointerException
        at __randomizedtesting.SeedInfo.seed([70B46EE4D84B4F3E:433D752FE0D7ED88]:0)
        at java.base/java.time.Instant.compareTo(Instant.java:1255)
        at org.opensearch.indexmanagement.transform.TransformRunnerIT$test no-op execution when no buckets have been modified$secondIterationMetadata$1.invoke(TransformRunnerIT.kt:428)
        at org.opensearch.indexmanagement.transform.TransformRunnerIT$test no-op execution when no buckets have been modified$secondIterationMetadata$1.invoke(TransformRunnerIT.kt:30)
        at org.opensearch.indexmanagement.TestHelpersKt.waitFor(TestHelpers.kt:119)
        at org.opensearch.indexmanagement.TestHelpersKt.waitFor$default(TestHelpers.kt:113)
        at org.opensearch.indexmanagement.transform.TransformRunnerIT.test no-op execution when no buckets have been modified(TransformRunnerIT.kt:424)
```
and 
```
REPRODUCE WITH: ./gradlew ':integTest' --tests "org.opensearch.indexmanagement.indexstatemanagement.action.RolloverActionIT.test rollover min primary shard size" -Dtests.seed=42BE7170D8C1087A -Dtests.security.manager=false -Dtests.locale=lv-LV -Dtests.timezone=Antarctica/Macquarie -Druntime.java=14
  2> java.lang.AssertionError: Found multiple shards over 100kb
        at __randomizedtesting.SeedInfo.seed([42BE7170D8C1087A:47E2B4CD4392AF2B]:0)
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.assertTrue(Assert.java:42)
        at org.opensearch.indexmanagement.indexstatemanagement.action.RolloverActionIT.test rollover min primary shard size(RolloverActionIT.kt:288)
```

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
```diff
@@ -390,6 +395,24 @@ class TransportAddPolicyAction @Inject constructor(
         private fun onFailure(t: Exception) {
             actionListener.onFailure(ExceptionsHelper.unwrapCause(t) as Exception)
         }
+
+        fun removeClusterStateMetadatas(indices: List<Index>) {
```
dbbaughe, line 399: nit: Seems like this is duplicated in the TransportRemovePolicyAction, could we just have a single helper fn to use in both?

dbbaughe (approved):

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
downsrob: Sounds good, I am changing this to a coroutine in the rest apis PR. I am going to leave it as is here and then refactor over there. LMK if you want it in this PR instead

bowenlan-amzn (approved):

PR #278, opened 2022-02-25
introducing custom uuid setting in SPI

thalurur:

*Issue #, if available:*
N/A

*Description of changes:*
* Allow extensions to set custom uuid index setting
* Enabled more tests
* Support extension status checker to determine if extension actions are supported 
* Clean up SPI interface

*CheckList:*


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/DefaultIndexMetadataService.kt
```diff
@@ -37,8 +37,11 @@ class DefaultIndexMetadataService : IndexMetadataService {
 
         response.state.metadata.indices.forEach {
             // TODO waiting to add document count until it is definitely needed
-            // TODO find a way to avoid this managed service code difference
-            val uuid = it.value.settings.get(COLD_UUID_SETTING, it.value.indexUUID) // use the cold uuid if it exists
+            val uuid = if (customUUIDSetting != null) {
+                it.value.settings.get(customUUIDSetting, it.value.indexUUID)
```
downsrob, line 41: Does `it.value.settings.get(null, it.value.indexUUID)` still return it.value.indexUUID, or does it throw an error? If it works then I would prefer to keep this as the one line setting check

spi/src/main/kotlin/org.opensearch.indexmanagement.spi/indexstatemanagement/StatusChecker.kt
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * com.maddyhome.idea.copyright.pattern.CommentInfo@2b5f8216
```
downsrob, line 5: Seeing this maddyhome thing again

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/DefaultIndexMetadataService.kt
thalurur: Unfortunately cannot pass null for the setting get key

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -322,9 +321,14 @@ object ManagedIndexRunner :
             }
         }
 
+        val actionExtensionName = ISMActionsParser.instance.customActionExtensionMap[action?.type]
         // If this action is not allowed and the step to be executed is the first step in the action then we will fail
         // as this action has been removed from the AllowList, but if its not the first step we will let it finish as it's already inflight
-        if (!action?.isAllowed(allowList) && action.isFirstStep(step?.name) && action.type != TransitionsAction.name) {
+        // We check if the action is allowed in cluster settings and the extension from which action is registered is enabled
+        if (action?.isAllowed(allowList) == false && step != null && action.isFirstStep(step.name) &&
+            action.type != TransitionsAction.name && !extensionStatusChecker.isEnabled(actionExtensionName)
```
downsrob, line 329: This would prevent us from failing any actions from enabled extensions right? To me the logic should be 
```
(action?.isAllowed(allowList) == false && step != null && action.isFirstStep(step.name) &&
            action.type != TransitionsAction.name) || !extensionStatusChecker.isEnabled(actionExtensionName)
```

thalurur: right, good point

downsrob: Hmm, but still this change breaks the allow list right? As it would mean that any default action cannot be disabled by removing it from the allow list?
How about
```
(action?.isAllowed(allowList) == false || !extensionStatusChecker.isEnabled(actionExtensionName)) && 
step != null && action.isFirstStep(step.name) && action.type != TransitionsAction.name
```

thalurur: makes sense changed

downsrob (approved):

spi/src/main/kotlin/org.opensearch.indexmanagement.spi/indexstatemanagement/model/ManagedIndexMetaData.kt
```diff
@@ -113,10 +113,9 @@ data class ManagedIndexMetaData(
         if (policyPrimaryTerm != null) builder.field(POLICY_PRIMARY_TERM, policyPrimaryTerm)
 
         // Only show rolled_over if we have rolled over or we are in the rollover action
-        // TODO: Fix this
-        /*if (rolledOver == true || (actionMetaData != null && actionMetaData.name == ActionConfig.ActionType.ROLLOVER.type)) {
+        if (rolledOver == true || (actionMetaData != null && actionMetaData.name == "rollover")) {
```
dbbaughe, line 116: Any specific reason we went from the actions type to the hardcoded "rollover" string?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -322,10 +321,15 @@ object ManagedIndexRunner :
...
         // If this action is not allowed and the step to be executed is the first step in the action then we will fail
         // as this action has been removed from the AllowList, but if its not the first step we will let it finish as it's already inflight
-        if (!action?.isAllowed(allowList) && action.isFirstStep(step?.name) && action.type != TransitionsAction.name) {
-            val info = mapOf("message" to "Attempted to execute action=${action.type} which is not allowed.")
+        // We check if the action is allowed in cluster settings and the extension from which action is registered is enabled
+        if ((action?.isAllowed(allowList) == false || !extensionStatusChecker.isEnabled(actionExtensionName)) && step != null &&
+            action.isFirstStep(step.name) && action.type != TransitionsAction.name
+        ) {
+
+            val info = mapOf("message" to "Attempted to execute action=${action?.type} which is not allowed.")
```
dbbaughe, line 332: Do you want to combine the check of allowList and extension enabled?
Looks like we could have left this part alone and then just added a completely separate check for if the extension is enabled and if not return a more explicit message like "Attempted to execute action [...] which is not enabled" or something.

Also not sure this is the behavior we want.. we chose to not completely stop an action if it's already executed the first step for the allow list because there could be unintended side effects if we don't allow it to finish, such as shrink allocating shards to a node and then stopping it and those shards are pinned there.

For an extension plugin saying it's not enabled anymore... even if it's already executed the first step that shouldn't matter.. because it'll fail (supposedly, as I think the intended use case for isEnabled essentially means will this even work) when trying to execute. If that's the case then for this extensions enabled check, it might make more sense to gracefully disable and tell them explicitly that the extension is not enabled anymore.

spi/src/main/kotlin/org.opensearch.indexmanagement.spi/indexstatemanagement/model/ManagedIndexMetaData.kt
downsrob: Because the ManagedIndexMetaData was moved to the spi it no longer has access to the index management action constants

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
downsrob: I separated the allowList check and the extension enabled check, now if an extension is not enabled then the extension's action will fail no matter what step it is on. I also updated some of the comments to reflect that an extension being enabled/disabled means it is on or off, and it does not represent extension health

downsrob (approved):

dbbaughe (approved):

Issue #200, opened 2021-11-17
Default CI Java Version to Java 11, run tests on 8, 14 and 17


dblock:
Coming from https://github.com/opensearch-project/opensearch-plugins/issues/64, default CI Java version to 11 (LTS), and run tests on 8, 14 and 17.

- [x] Change the default Java version for CI in GitHub actions from 14/15 to 11 (LTS).
- [ ] Add Java 8, 14 (used before), and 17 (next LTS) to the CI matrix.
- [x] Document Java versions in DEVELOPER_GUIDE.md.

dblock:
@downsrob I believe you only have JDK8 left to do for 1.3.0 here.

downsrob:
Just built with JDK8 and got a compilation failure
```
e: /home/runner/work/index-management/index-management/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/forcemerge/WaitForForceMergeStep.kt: (62, 56): Type mismatch: inferred type is Long but BigDecimal! was expected
```
@dblock Do we need build support for JDK 8 in the 1.3 release? As in we will need to get this fixed on the 1.3 branch by the end of the day?

downsrob:
There were a couple of small lines to change to get JDK 8 building. Not sure if this is something I should backport after the code freeze date, so I just have it in a draft PR here https://github.com/opensearch-project/index-management/pull/297

PR #284, opened 2022-03-04
Changes integ test java version from 14 to 11

downsrob:

*Issue #, if available:*
https://github.com/opensearch-project/index-management/issues/200

*Description of changes:*
Changes the default Java version for CI in GitHub actions from 14 to 11. 
*CheckList:*

### DIFF
--- a/.github/workflows/multi-node-test-workflow.yml
+++ b/.github/workflows/multi-node-test-workflow.yml
@@ -16,10 +16,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       # This step uses the setup-java Github action: https://github.com/actions/setup-java
-      - name: Set Up JDK 14
+      - name: Set Up JDK 11
         uses: actions/setup-java@v1
         with:
-          java-version: 14
+          java-version: 11
       # index-management
       - name: Checkout Branch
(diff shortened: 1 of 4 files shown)


dbbaughe:
Is there a reason for going down? And any reason for not having both run in a matrix if both are supported?

downsrob:
> Is there a reason for going down? And any reason for not having both run in a matrix if both are supported?

@dbbaughe core java version is going to 11 so we are being asked to change CI to the same version. Core is going down as 14 is not a long term support java version. We were asked to include java 8, 11, 14, and 17 in a matrix (ref: https://github.com/opensearch-project/index-management/issues/200)
I didn't do that because we have flaky test failures again. Say we made the multinode test run a matrix of 8, 11, 14, and 17. If we have a 15% chance of having a flaky failure on any given run, then we are failing at least one of the runs nearly 50% of the time (1 - .85^4 = .48), and when you retry, you can't retry just one version, you need to retry the full matrix.

Let me know if you think we still should though, or if you think we should do a matrix of just two versions. Java 17 is what we will be moving to next.

bowenlan-amzn (approved):

PR #286, opened 2022-03-07
Adds missed flags

downsrob:
Signed-off-by: Clay Downs <89109232+downsrob@users.noreply.github.com>

*Issue #, if available:*
NA

*Description of changes:*
Ports over some additional flags which seem to have been missed from previous bug fixes.

*CheckList:*


dbbaughe:
Can you explain what the changes do and if they were required why didn't any of the tests fail 🤔

downsrob:
The 'runTimeCounter = 0' line seemed to be missed in the metadata service backports. The purpose of setting the flag back to 0 would be that after the migration succeeds and the finishFlag is set to true, we reset the counter so that if it is reenabled it doesn't carry over the previous retry count. We don't have any tests which cover this so it didn't fail.
For the .routing(uuid) flag, we try to route all metadata requests by uuid, and if you use routing you need to use it always with getting, deleting or updating the document. It seems like it could cause some test flakiness in multi node tests but we don't have any tests which would directly fail because of it

bowenlan-amzn (approved):

dbbaughe (approved):

PR #323, opened 2022-04-01
Make sure qualifier default is alpha1 in IM

peterzhuamazon:
Signed-off-by: Peter Zhu <zhujiaxi@amazon.com>

*Issue #, if available:*
#303

*Description of changes:*
Make sure qualifier default is alpha1 in IM

*CheckList:*

### DIFF
--- a/.github/workflows/multi-node-test-workflow.yml
+++ b/.github/workflows/multi-node-test-workflow.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Checkout Branch
         uses: actions/checkout@v2
       - name: Run integration tests with multi node config
-        run: ./gradlew integTest -PnumNodes=3 -Dopensearch.version=2.0.0-alpha1-SNAPSHOT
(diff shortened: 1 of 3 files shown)


dbbaughe:
I'd imagine this gets tiring to update all repos everywhere all the time with these small changes.
Have we considered moving this type of logic into something like a gradle plugin that exists in one place that all plugins use?
So you could just update it there and have it automatically consumed by plugins (I think gradle can support ranges, at least form 7 onwards)

dbbaughe (approved):

downsrob (approved):

peterzhuamazon:
> I'd imagine this gets tiring to update all repos everywhere all the time with these small changes. Have we considered moving this type of logic into something like a gradle plugin that exists in one place that all plugins use? So you could just update it there and have it automatically consumed by plugins (I think gradle can support ranges, at least form 7 onwards)

That is a good suggestion we can definitely do this.
I will write a issue on this for later improvements.
Thanks.

Issue #305, opened 2022-03-14
[BUG] Replace checked-in ZIPs with dynamic dependencies 


dblock:
**Describe the bug**

Coming form https://github.com/opensearch-project/opensearch-build/issues/1750, remove check-in ZIPs and replace them with dynamic dependencies.

```
./index-management/src/test/resources/bwc/job-scheduler/1.13.0.0/opendistro-job-scheduler-1.13.0.0.zip
./index-management/src/test/resources/bwc/indexmanagement/1.13.2.0/opendistro-indexmanagement-1.13.2.0.zip
./index-management/src/test/resources/job-scheduler/opensearch-job-scheduler-1.3.0.0-SNAPSHOT.zip
./index-management/src/test/resources/lang-mustache/lang-mustache-1.0.0.zip
```

The first 3 can be downloaded from stable or `latest` URIs. 
Lang-mustache needs to be replaced with an external dependency.

PR #327, opened 2022-04-12
Replace checked-in ZIPs with dynamic dependencies

downsrob:

*Issue #, if available:*
https://github.com/opensearch-project/index-management/issues/305

*Description of changes:*
Replace checked-in ZIPs with dynamic dependencies. The current job scheduler zip used for integration tests is downloaded from ci.opensearch.org. The backwards compatibility test job scheduler and index management versions are downloaded from https://github.com/opendistro-for-elasticsearch artifacts. These dependencies are downloaded during the gradle configuration phase, and will be refreshed with any task run.

Removes the mustache zip entirely, as I couldn't find any reference to it being used.

*CheckList:*


build.gradle
```diff
@@ -234,6 +240,17 @@ test {
...
+ext.getPluginResource = { download_to_folder, download_from_src ->
+    project.mkdir download_to_folder
+    ant.get(src: download_from_src,
+            dest: download_to_folder,
+            httpusecaches: false)
+    return fileTree(download_to_folder).getSingleFile()
+}
+
+// Download the job scheduler test dependency
+getPluginResource(job_scheduler_resource_folder, job_scheduler_build_download)
```
thalurur, line 252: by calling this at this level in build.gradle mean this will get executed always for all targets?

downsrob: That's right. I believe that we also initialize the integTest clusters with the plugin zips for all targets in the configuration phase, so to get around downloading the resource always, we would need to refactor the cluster setup, which I didn't have time to take on at this point

thalurur: got it, lets create an issue for this - we don't need to address it immediately

thalurur (approved):

bowenlan-amzn (approved):

build.gradle
downsrob: https://github.com/opensearch-project/index-management/issues/332

PR #328, opened 2022-04-13
Fix metadata migration logic error when update setting call failed

bowenlan-amzn:
If update setting to 1 failed, `runTimeCounter` shouldn't be increased.
Otherwise it will increase to be larger than 10, and start to update
setting to -1 in the end which is unwanted.

*Issue #, if available:*

*Description of changes:*

*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/MetadataService.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/MetadataService.kt
@@ -56,7 +56,8 @@ class MetadataService(
     private var failedToCleanIndices = mutableSetOf<Index>()
 
     private var counter = 0
-    private var runTimeCounter = 0
+    final var runTimeCounter = 1
+        private set
     private val maxRunTime = 10
 
     // used in coordinator sweep to cancel scheduled process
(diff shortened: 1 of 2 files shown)


thalurur:
Can include the exists check in template migration - https://github.com/opensearch-project/index-management/blob/b6e7d35c4fe4b64a41ef9e45d0cafcdfe0289042/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/migration/MigrationServices.kt#L80

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/MetadataService.kt
```diff
@@ -84,12 +85,18 @@ class MetadataService(
                 return
             }
 
-            if (runTimeCounter >= maxRunTime) {
+            if (!imIndices.indexManagementIndexExists()) {
+                logger.info("ISM config index not exist, so we cancel the metadata migration job.")
+                finishFlag = true; runningLock = false; runTimeCounter = 0
+                return
+            }
```
thalurur, line 92: so if there is no IM config index then we don't update the setting to 1 - is my understanding correct?

bowenlan-amzn:
> Can we also include the exists check in template migration? -
> 
> https://github.com/opensearch-project/index-management/blob/b6e7d35c4fe4b64a41ef9e45d0cafcdfe0289042/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/migration/MigrationServices.kt#L80

Since template migration migrates from index template, config index not existing doesn't really mean there is no related index template, so I didn't add this check this time.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/MetadataService.kt
thalurur: Can we just check in the initMoveMetadata in that case to not even start the metadata migration if there is no IM config index? https://github.com/opensearch-project/index-management/blob/b6e7d35c4fe4b64a41ef9e45d0cafcdfe0289042/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexCoordinator.kt#L505

bowenlan-amzn: The only reason I have is logically I think this check belongs to the migration logic. But either way is fine if you think we should do it outside.

thalurur: I slightly prefer being outside but this works too - but this is fine too

thalurur (approved):

downsrob (approved):

Issue #333, opened 2022-04-14
[BUG][DISTRIBUTION][2.0.0] IM downloading JS zip that is not exist due to no success builds


peterzhuamazon:
```
+ ./gradlew assemble --no-daemon --refresh-dependencies -DskipTests=true -Dopensearch.version=2.0.0-rc1-SNAPSHOT -Dbuild.snapshot=true -Dbuild.version_qualifier=rc1
To honour the JVM settings for this build a single-use Daemon process will be forked. See https://docs.gradle.org/7.3.3/userguide/gradle_daemon.html#sec:disabling_the_daemon.
Daemon will be stopped at the end of the build 

FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* What went wrong:
A problem occurred evaluating root project 'opensearch-index-management'.
> Can't get https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.0.0-rc1/latest/linux/x64/builds/opensearch/plugins/opensearch-job-scheduler-2.0.0.0-rc1.zip to /tmp/tmpbpozmgws/index-management/src/test/resources/job-scheduler/opensearch-job-scheduler-2.0.0.0-rc1.zip

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
==============================================================================

2: Task failed with an exception.
-----------
* Where:
Build file '/tmp/tmpbpozmgws/index-management/build.gradle' line: 208

* What went wrong:
A problem occurred configuring root project 'opensearch-index-management'.
> Index 0 out of bounds for length 0

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
==============================================================================

* Get more help at https://help.gradle.org/

Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

See https://docs.gradle.org/7.3.3/userguide/command_line_interface.html#sec:command_line_warnings

BUILD FAILED in 10s
2022-04-14 22:20:11 ERROR    Error building index-management, retry with: ./build.sh manifests/2.0.0/opensearch-2.0.0.yml --component index-management --snapshot
Traceback (most recent call last):
  File "./src/run_build.py", line 80, in <module>
    sys.exit(main())
[...]

PR #334, opened 2022-04-15
Only download JS zip when integTest is running

peterzhuamazon:
Signed-off-by: Peter Zhu <zhujiaxi@amazon.com>

*Issue #, if available:*
#333

*Description of changes:*
Only download JS zip when integTest is running

*CheckList:*

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -378,11 +378,15 @@ String bwcFilePath = "src/test/resources/bwc/"
 String bwc_js_resource_location = bwcFilePath + "job-scheduler/" + bwcJobSchedulerVersion
 String bwc_im_resource_location = bwcFilePath + "indexmanagement/" + bwcVersion
 
+// Downloads the bwc job scheduler version
 String bwc_js_download_url = "https://github.com/opendistro-for-elasticsearch/job-scheduler/releases/download/v" +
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -268,10 +265,23 @@ testClusters.integTest {
             debugPort += 1
         }
     }
-    plugin(provider({
-        new RegularFile() {
-            @Override
-            File getAsFile() { fileTree(job_scheduler_resource_folder).getSingleFile() }
+
+    plugin(provider(new Callable<RegularFile>(){
```
thalurur, line 269: out of curiosity: why not just call the getPluginResource before trying to load the file?

peterzhuamazon: Tried to do that initially but the call still happens during assemble.

peterzhuamazon:
Interesting that bwc test success on my local while instantly fail in github actions.

build.gradle
peterzhuamazon: Have some research and figure out the correct location now.
Thanks.

peterzhuamazon: Some more testing and that method I thought runs didnt.
Revert back.

dblock (approved):

downsrob (approved):

PR #335, opened 2022-04-18
Fixes shrink action shard allocation checks

downsrob:

*Issue #, if available:*
NA

*Description of changes:*
There were a few issues with Shrink shard allocation which would cause failures when shrinking indices with replicas:
- During shard allocation dry run, a move shard request was initiated for every shard not present on the target node. This could mean a primary shard and multiple replicas. When multiple requests for the same shard are initiated at one time, the dry run will fail with `explanation=a copy of this shard is already allocated to this node`. To fix this, this PR only creates an allocation dry run request for one random shard copy for each shard.
- During shard allocation dry run, allocation requests were still created for shards which had a copy on the target node already, because of this the dry run would fail with `explanation=a copy of this shard is already allocated to this node`. To fix this, this PR checks if any copy of the shard is present on the target node to initiate a request.
- While waiting for shard allocation to finish, we previously only checked that primary shards were on the target node, meaning that a replica shard being on the target node instead would not allow shrink. This PR fixes that behavior by checking if any copy of each shard is on the target node and started.

*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/WaitForMoveShardsStep.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/WaitForMoveShardsStep.kt
@@ -63,16 +63,16 @@ class WaitForMoveShardsStep(private val action: ShrinkAction) : Step(name) {
             val nodeToMoveOnto = localShrinkActionProperties.nodeName
             val inSyncAllocations = context.clusterService.state().metadata.indices[indexName].inSyncAllocationIds
             val numReplicas = context.clusterService.state().metadata.indices[indexName].numberOfReplicas
-            var numShardsOnNode = 0
+            val shardIdOnNode: MutableMap<Int, Boolean> = mutableMapOf()
             var numShardsInSync = 0
             for (shard: ShardStats in response.shards) {
                 val routingInfo = shard.shardRouting
                 val nodeIdShardIsOn = routingInfo.currentNodeId()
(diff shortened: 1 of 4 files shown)


thalurur (approved):

downsrob:
Test and build workflow CI failure was due to unrelated flaky test

bowenlan-amzn (approved):

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/shrink/WaitForMoveShardsStep.kt
```diff
@@ -63,28 +63,28 @@ class WaitForMoveShardsStep(private val action: ShrinkAction) : Step(name) {
...
-            var numShardsOnNode = 0
+            val shardIdOnNode: MutableMap<Int, Boolean> = mutableMapOf()
             var numShardsInSync = 0
             for (shard: ShardStats in response.shards) {
                 val routingInfo = shard.shardRouting
                 val nodeIdShardIsOn = routingInfo.currentNodeId()
                 val nodeNameShardIsOn = context.clusterService.state().nodes()[nodeIdShardIsOn].name
+                if (nodeNameShardIsOn.equals(nodeToMoveOnto) && routingInfo.started()) {
+                    shardIdOnNode[shard.shardRouting.id] = true
+                }
```
bowenlan-amzn, line 74: Put a note here, we want to extract below `if (routingInfo.primary())` block into a function `checkReplicaInSync` for readability.

PR #339, opened 2022-04-19
BugFix: Notification integration issues

thalurur:

*Issue #, if available:*
N/A

*Description of changes:*
At the moment when notification plugin is called from IM the thread context is not populated with user information as a result when filter by backend roles is enabled in notification - IM won't be able to successfully send the notification. Updating logic to populate the thread context correctly

Added all the required run time dependencies to successfully send notifications through custom webhook

*CheckList:*

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -174,6 +174,8 @@ dependencies {
     implementation "org.opensearch:common-utils:${common_utils_version}"
     implementation "com.github.seancfoley:ipaddress:5.3.3"
     implementation "commons-codec:commons-codec:${versions.commonscodec}"
+    implementation "org.apache.httpcomponents:httpclient:4.5.13"
+    implementation "org.apache.httpcomponents:httpcore:4.4.12"
 
     testImplementation "org.opensearch.test:framework:${opensearch_version}"
     testImplementation "org.jetbrains.kotlin:kotlin-test:${kotlin_version}"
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -174,6 +174,8 @@ dependencies {
     implementation "org.opensearch:common-utils:${common_utils_version}"
     implementation "com.github.seancfoley:ipaddress:5.3.3"
     implementation "commons-codec:commons-codec:${versions.commonscodec}"
+    implementation "org.apache.httpcomponents:httpclient:4.5.13"
```
bowenlan-amzn, line 177: Curious why these are added along with this change?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/util/NotificationUtils.kt
```diff
@@ -45,18 +47,37 @@ suspend fun LegacyBaseMessage.publishLegacyNotification(client: Client) {
...
-    val res: SendNotificationResponse = NotificationsPluginInterface.suspendUntil {
-        this.sendNotification(
-            (client as NodeClient),
-            managedIndexMetaData.getEventSource(title),
-            ChannelMessage(compiledMessage, null, null),
-            listOf(channel.id),
-            it
-        )
+    client.threadPool().threadContext.stashContext().use {
+        // We need to set the user context information in the thread context for notification plugin to correctly resolve the user object
```
dbbaughe, line 59: Are we setting the user object twice in the threadContext? Here and above the calling stack in the runner?

dbbaughe (approved):

downsrob (approved):

build.gradle
thalurur: It seems these are needed for the sending notification to legacy custom_webhook endpoints, the libraries are no longer in class path - I am not entirely sure where this was removed. Since IM needs it bundling it as part of the zip. 

The PR is meant to fix bugs with notification plugin integration and since this is one of the bugs fixed it - I added a line in overview for this. However, I think I missed the renaming of PR - I updated it now

PR #340, opened 2022-04-20
Fixes flaky continuous transforms and shrink tests

downsrob:

*Issue #, if available:*
NA

*Description of changes:*
- Fixes two continuous transforms flaky tests which were not waiting until the metadata was fully updated
- Fixes some of the shrink ITs flakiness. These tests would sometimes fail on multi node setups because of the job scheduler index being moved between nodes. When the index was moved, all jobs would be descheduled so the force start window could be missed.
- Shrink ITs were failing on remote tests against the release docker image as the docker images did not have enough free space to shrink an index and stay below the watermark thresholds. To fix this, now the shrink ITs set the watermark settings to be just a few bytes before each test.
- When fixing the shrink ITs I noticed that the settings object passed in from the step context did not contain the watermark levels, so the default was always used. This was fixed by building a new settings object from the cluster settings to get these settings directly from the cluster state.
- Clears indices and data streams after transforms and rollups test suites finish. Previously if the TransformRunnerITs or RollupRunnerIts were run twice in a row against the same remote cluster, the indices from the first run would stick around, messing up the second run.

*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/util/StepUtils.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/util/StepUtils.kt
@@ -16,6 +16,9 @@ import org.opensearch.client.Client
 import org.opensearch.cluster.metadata.IndexMetadata
 import org.opensearch.cluster.node.DiscoveryNodes
 import org.opensearch.cluster.routing.allocation.DiskThresholdSettings
+import org.opensearch.cluster.routing.allocation.DiskThresholdSettings.CLUSTER_ROUTING_ALLOCATION_DISK_FLOOD_STAGE_WATERMARK_SETTING
+import org.opensearch.cluster.routing.allocation.DiskThresholdSettings.CLUSTER_ROUTING_ALLOCATION_HIGH_DISK_WATERMARK_SETTING
+import org.opensearch.cluster.routing.allocation.DiskThresholdSettings.CLUSTER_ROUTING_ALLOCATION_LOW_DISK_WATERMARK_SETTING
 import org.opensearch.common.settings.ClusterSettings
 import org.opensearch.common.settings.Settings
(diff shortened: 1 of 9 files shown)


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/util/StepUtils.kt
```diff
@@ -149,16 +152,29 @@ fun getFreeBytesThresholdHigh(settings: Settings, clusterSettings: ClusterSettin
     } else diskThresholdBytes.bytes
 }
 
+fun getDiskSettings(clusterSettings: ClusterSettings): Settings {
+    return Settings.builder().put(
```
dbbaughe, line 156: What's the purpose of this (creating this almost empty Settings instance instead of passing the one that presumably has all the settings)?

bowenlan-amzn: I'm wondering the reason to do this is previously the settings passed in to `DiskThresholdSettings` are not right? The right way is to extract from clusterSettings?

bowenlan-amzn: Makes sense, I agree we need to remove settings from step context and only use clusterSettings to get update-to-date setting values.

bowenlan-amzn (approved):

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/util/StepUtils.kt
dbbaughe: Got it, maybe @thalurur has some input on it or it might have just been a miss that we were passing in a snapshot of the settings.

dbbaughe (approved):

PR #366, opened 2022-05-17
Uses custom plugin to publish zips to maven

downsrob:

*Issue #, if available:*
https://github.com/opensearch-project/index-management/issues/348

*Description of changes:*
Adds the custom plugin to publish zips to maven to the build.gradle. Introduces a custom build.sh script.

*CheckList:*

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -69,7 +69,6 @@ fi
 mkdir -p $OUTPUT
 
 ./gradlew assemble --no-daemon --refresh-dependencies -DskipTests=true -Dopensearch.version=$VERSION -Dbuild.snapshot=$SNAPSHOT -Dbuild.version_qualifier=$QUALIFIER
-./gradlew publishPluginZipPublicationToZipStagingRepository -Dopensearch.version=$VERSION -Dbuild.snapshot=$SNAPSHOT -Dbuild.version_qualifier=$QUALIFIER
 
 zipPath=$(find . -path \*build/distributions/*.zip)
 distributions="$(dirname "${zipPath}")"
@@ -77,3 +76,7 @@ distributions="$(dirname "${zipPath}")"
 echo "COPY ${distributions}/*.zip"
 mkdir -p $OUTPUT/plugins
 cp ${distributions}/*.zip ./$OUTPUT/plugins
+
+./gradlew publishPluginZipPublicationToZipStagingRepository -Dopensearch.version=$VERSION -Dbuild.snapshot=$SNAPSHOT -Dbuild.version_qualifier=$QUALIFIER
+mkdir -p $OUTPUT/maven/org/opensearch
+cp -r ./build/local-staging-repo/org/opensearch/. $OUTPUT/maven/org/opensearch
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+
+#
+# Copyright OpenSearch Contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -ex
+
+function usage() {
+    echo "Usage: $0 [args]"
+    echo ""
+    echo "Arguments:"
+    echo -e "-v VERSION\t[Required] OpenSearch version."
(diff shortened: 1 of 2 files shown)


downsrob:
@thalurur For BWC testing we utilize a zip. The only use case I can see for the leaves would be if we hope to utilize the maven zips in the future for our BWC testing

downsrob:
> Wondering do we need to exclude task in settings.gradle like [here](https://github.com/opensearch-project/job-scheduler/pull/182/files#diff-7f825392aa37acd1cee0c2e7b9bb7366ad6eac64f3e6cdd816e156bcb69d30deR13)

I believe that we do not. According to step 7 in the [setup guide](https://github.com/opensearch-project/opensearch-plugins/blob/main/BUILDING.md#opensearchpluginzip) for this plugin, "Exclude the following tasks in settings.gradle file, if the build script exists for the plugin and has the tasks publishToMavenLocal and publishAllPublicationsToStagingRepository, ..."

bowenlan-amzn (dismissed):

dblock:
cc: @prudhvigodithi

That [link](https://github.com/opensearch-project/job-scheduler/pull/182/files#diff-7f825392aa37acd1cee0c2e7b9bb7366ad6eac64f3e6cdd816e156bcb69d30deR13) in job-scheduler looks suspicious, care to track it down? not sure why that was needed

prudhvigodithi:
> That [link](https://github.com/opensearch-project/job-scheduler/pull/182/files#diff-7f825392aa37acd1cee0c2e7b9bb7366ad6eac64f3e6cdd816e156bcb69d30deR13) in job-scheduler looks suspicious, care to track it down? not sure why that was needed

Hey @dblock from [document](https://github.com/opensearch-project/opensearch-plugins/blob/main/BUILDING.md#opensearchpluginzip) point 7, this targets `ALL` publications for local maven repo and staging maven repo that includes `publishPluginZipPublicationToMavenLocal, publishPluginZipPublicationToStagingRepository`, I mean its fine to allow it to run as well without `excludedTaskNames`, should work but it will push the zips twice as `publishPluginZipPublicationToStagingRepository` will be called before `publishPluginZipPublicationToZipStagingRepository` which is the actual publish task.
Screen Shot 2022-05-24 at 6 31 45 PM

scripts/build.sh
```diff
@@ -0,0 +1,79 @@
...
+
+./gradlew assemble --no-daemon --refresh-dependencies -DskipTests=true -Dopensearch.version=$VERSION -Dbuild.snapshot=$SNAPSHOT -Dbuild.version_qualifier=$QUALIFIER
+./gradlew publishPluginZipPublicationToZipStagingRepository -Dopensearch.version=$VERSION -Dbuild.snapshot=$SNAPSHOT -Dbuild.version_qualifier=$QUALIFIER
+
+zipPath=$(find . -path \*build/distributions/*.zip)
+distributions="$(dirname "${zipPath}")"
+
+echo "COPY ${distributions}/*.zip"
+mkdir -p $OUTPUT/plugins
+cp ${distributions}/*.zip ./$OUTPUT/plugins
```
prudhvigodithi, line 79: We should be also adding 
```
mkdir -p $OUTPUT/maven/org/opensearch
cp -r ./build/local-staging-repo/org/opensearch/. $OUTPUT/maven/org/opensearch 
```
As this will put the local staged maven artifacts under `$OUTPUT/maven/org/opensearch ` and CI will be using this folder `$OUTPUT/maven/org/opensearch ` to publish to end maven 
Please refer to `job-scheduler` https://github.com/opensearch-project/job-scheduler/blob/main/scripts/build.sh#L77-L79

downsrob: Thanks! Added.

prudhvigodithi: Looks good.
Please try to run the the build component locally, once passed `tar/builds/opensearch/maven/org/opensearch/plugin/opensearch-index-management/` folder should be created and zip should be added as per maven coordinates.
Thank you

prudhvigodithi (approved):

bowenlan-amzn (approved):

AWSHurneyt (approved):

PR #371, opened 2022-05-23
BWC for rollover skip, restricted index pattern

bowenlan-amzn:

*Description of changes:*

Though these 2 settings are released after OpenSearch, we backport them to AWS Elasticsearch code base, to keep the minimal code difference, we want to add the change for supporting setting BWC between Elasticsearch and OpenSearch code bases.

*CheckList:*


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/opensearchapi/OpenSearchExtensions.kt
```diff
@@ -55,6 +55,9 @@ fun IndexMetadata.getRolloverAlias(): String? {
 }
 
 fun IndexMetadata.getRolloverSkip(): Boolean {
+    if (this.settings.get(ManagedIndexSettings.ROLLOVER_SKIP.key).isNullOrBlank()) {
+        return this.settings.getAsBoolean(LegacyOpenDistroManagedIndexSettings.ROLLOVER_SKIP.key, false)
+    }
```
downsrob, line 60: I believe this is unnecessary because the fallback setting for ManagedIndexSettings.ROLLOVER_SKIP is set to LegacyOpenDistroManagedIndexSettings.ROLLOVER_SKIP when initialized, so the open distro setting would be checked anyways

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/settings/LegacyOpenDistroManagedIndexSettings.kt
```diff
@@ -191,5 +199,13 @@ class LegacyOpenDistroManagedIndexSettings {
...
             Setting.Property.Deprecated
         )
+
+        val RESTRICTED_INDEX_PATTERN = Setting.simpleString(
+            "opendistro.index_state_management.restricted_index_pattern",
+            ManagedIndexSettings.DEFAULT_RESTRICTED_PATTERN,
+            Setting.Property.NodeScope,
+            Setting.Property.Dynamic,
+            Setting.Property.Deprecated
+        )
```
thalurur, line 209: Thanks for adding this too Bowen

thalurur (approved):

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/opensearchapi/OpenSearchExtensions.kt
bowenlan-amzn: Based on above `getRolloverAlias`, probably this is not working as that. Write like this should be safe.

downsrob (approved):

PR #419, opened 2022-07-19
removed recursion from Explain Action to avoid stackoverflow in some …

petardz:
…situations

Signed-off-by: Petar Dzepina <petar.dzepina@gmail.com>

*Issue #, if available:* [410](https://github.com/opensearch-project/index-management/issues/410)

*Description of changes:*
Removed recursion from Explain Action to avoid stack overflow bug when there's few hundreds+ managed indexes present
*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/explain/TransportExplainAction.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/explain/TransportExplainAction.kt
@@ -52,6 +52,7 @@ import org.opensearch.indexmanagement.indexstatemanagement.util.MetadataCheck
 import org.opensearch.indexmanagement.indexstatemanagement.util.checkMetadata
 import org.opensearch.indexmanagement.indexstatemanagement.util.managedIndexMetadataID
 import org.opensearch.indexmanagement.opensearchapi.parseWithType
+import org.opensearch.indexmanagement.opensearchapi.suspendUntil
 import org.opensearch.indexmanagement.spi.indexstatemanagement.model.ISMIndexMetadata
 import org.opensearch.indexmanagement.spi.indexstatemanagement.model.ManagedIndexMetaData
 import org.opensearch.indexmanagement.util.SecurityUtils.Companion.buildUser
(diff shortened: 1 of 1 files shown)


downsrob:
Looks great Petar, thanks!

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/explain/TransportExplainAction.kt
```diff
@@ -339,68 +340,29 @@ class TransportExplainAction @Inject constructor(
...
-                            false -> {
-                                actionListener.onFailure(e)
-                            }
-                        }
+            CoroutineScope(Dispatchers.IO).launch {
+                // filter out indicies for which user doesn't have manage index permissions
+                for (i in 0 until indexNames.count()) {
+                    val request = ManagedIndexRequest().indices(indexNames[i])
+                    try {
+                        client.suspendUntil<NodeClient, AcknowledgedResponse> { execute(ManagedIndexAction.INSTANCE, request, it) }
```
getsaurabh02, line 349: Although this will solve the stack overflow issue, I still see the multiple calls for ``TransportManagedIndexAction`` for every index as a potential problem and bottleneck for the clusters with large number of indices. It could still block the primary (transport) worker thread until all the transport action calls succeeds or fails for every index and could lead to starvation issue.

The non operational transport action that is used by ISM to check if the user has required index permissions to manage index in form of [TransportManagedIndexAction](https://github.com/opensearch-project/index-management/blob/c739190a9f807232cc500acd12bc1f57d5d37c7e/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/managedIndex/TransportManagedIndexAction.kt#L28-L30) seems to be a workaround due to lack of better ways for mapping user permission with indices. 

I will prefer instrumenting some cleaner and direct ways of identifying index permissions for the user. This is outside the scope of this PR, but we should open an issue and follow up separately on the possible approaches.

getsaurabh02 (approved):
Thanks for the changes. This looks much cleaner and simpler now. 
Also, lets follow up separately on a new issue to discuss the potential alternatives to the [TransportManagedIndexAction](https://github.com/opensearch-project/index-management/blob/c739190a9f807232cc500acd12bc1f57d5d37c7e/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/managedIndex/TransportManagedIndexAction.kt#L28-L30) as suggested in the previous comment.

praveensameneni (approved):
LGTM

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/explain/TransportExplainAction.kt
getsaurabh02: Created https://github.com/opensearch-project/index-management/issues/422 so that we don't miss out on this and look into this at priority.

cc: @qreshi @downsrob @petardz @praveensameneni

PR #428, opened 2022-07-25
Continuous Execution

ronnaksaxena:
Signed-off-by: Ronnak Saxena <ronsax@amazon.com>
*Issue #, if available:*

*Description of changes:*
•Added a continuous execution option when adding polices to managed indices
•Speeds up execution of indices that can run multiple actions/steps in a single execution cycle within runner logic
•Modified "Add Policy", "Update Managed Index Policy", & "Explain Index" APIs to incorporate continuous execution feature

*CheckList:*
- [ X] Commits are signed per the DCO using --signoff

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,8 +13,8 @@ import java.util.function.Predicate
 
 buildscript {
     ext {
-        isSnapshot = "true" == System.getProperty("build.snapshot", "true")
-        opensearch_version = System.getProperty("opensearch.version", "2.1.0-SNAPSHOT")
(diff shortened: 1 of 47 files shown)


downsrob:
Based on the feedback from the presentation of this feature, we would like a mechanism to support bulk migrating indices to continuous. The change policy API does support wildcards, but it requires a policyID in the request, so users can only bulk change indices to be continuous if the policies are set as well.

Before this feature is released, let's do a minor refactor to allow null policyID values in a ChangePolicyRequest if there is a continuous field present.

Issue #61, opened 2021-06-03
Allow a dynamic target index in ISM Rollup


aditjind:
<a href="https://github.com/mark-meyer"><img src="https://avatars.githubusercontent.com/u/7340303?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [mark-meyer](https://github.com/mark-meyer)**
_Wednesday Mar 31, 2021 at 00:46 GMT_
_Originally opened as https://github.com/opendistro-for-elasticsearch/index-management/issues/428_

----

I would like to apply a policy to daily log indices that performs a rollup in one of the states before deleting the index. However is seem the `target_index` requires a string and there is no option to interpolate something like `{{ctx.index}}` or similar. This makes it impossible to apply this policy to more than one index unless you are okay with rollup overwriting the previous one, which you probably aren't.    

**Describe the solution you'd like**
I would like to specify the `target_index` field of an ism rollup dynamically based on the name (or some other metadata) of the index being rolled up such as `"target_index": "rollup_{{ctx.index}}"`

aditjind:
<a href="https://github.com/dbbaughe"><img src="https://avatars.githubusercontent.com/u/46505179?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [dbbaughe](https://github.com/dbbaughe)**
_Wednesday Mar 31, 2021 at 00:51 GMT_

----

I believe @thalurur just did something similar for snapshot name too.

That being said, correct me if I'm wrong Ravi, but if @mark-meyer applies this policy to daily log indices even if they roll up into the same target index they will not overwrite each other. We do support multiple rollups in the same index. Each index that is managed by the policy would have a rollup job created for itself when the rollup action is created which will use its own rollup ID as a namespace in the target index to handle conflicts. And the rollup ID seems to consist of a hash of the job configuration and the source index which will be unique per daily log index.

Edit: Do we delete these temporary rollup jobs after they finish @thalurur?

aditjind:
<a href="https://github.com/thalurur"><img src="https://avatars.githubusercontent.com/u/6005951?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [thalurur](https://github.com/thalurur)**
_Monday Apr 05, 2021 at 22:30 GMT_

----

[...]

PR #435, opened 2022-07-28
added support for mustache scripting of rollup.target_index field

petardz:
*Issue #, if available:* [61](https://github.com/opensearch-project/index-management/issues/61)

*Description of changes:*

Added capability to use mustache scripting as target_index field value. Only supported param right now is source_index: target_index: rollup_{{ctx.source_index}}

*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/rollup/action/index/TransportIndexRollupAction.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/rollup/action/index/TransportIndexRollupAction.kt
@@ -92,6 +92,14 @@ class TransportIndexRollupAction @Inject constructor(
             if (response.isAcknowledged) {
                 log.info("Successfully created or updated $INDEX_MANAGEMENT_INDEX with newest mappings.")
                 if (request.opType() == DocWriteRequest.OpType.CREATE) {
+                    if (!validateTargetIndexName()) {
+                        return actionListener.onFailure(
+                            OpenSearchStatusException(
+                                "target_index value is invalid: ${request.rollup.targetIndex}",
+                                RestStatus.BAD_REQUEST
+                            )
+                        )
+                    }
                     putRollup()
                 } else {
                     getRollup()
--- a/src/main/kotlin/org/opensearch/indexmanagement/rollup/action/index/TransportIndexRollupAction.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/rollup/action/index/TransportIndexRollupAction.kt
@@ -28,6 +28,7 @@ import org.opensearch.commons.authuser.User
 import org.opensearch.indexmanagement.IndexManagementIndices
 import org.opensearch.indexmanagement.IndexManagementPlugin.Companion.INDEX_MANAGEMENT_INDEX
 import org.opensearch.indexmanagement.rollup.model.Rollup
+import org.opensearch.indexmanagement.rollup.util.RollupFieldValueExpressionResolver
 import org.opensearch.indexmanagement.rollup.util.parseRollup
 import org.opensearch.indexmanagement.settings.IndexManagementSettings
 import org.opensearch.indexmanagement.util.IndexUtils
@@ -143,6 +144,12 @@ class TransportIndexRollupAction @Inject constructor(
         }
 
         private fun putRollup() {
+            if (!validateTargetIndexName()) {
+                return actionListener.onFailure(OpenSearchStatusException(
+                    "target_index value is invalid: ${request.rollup.targetIndex}",
+                    RestStatus.BAD_REQUEST)
+                )
+            }
             val rollup = request.rollup.copy(schemaVersion = IndexUtils.indexManagementConfigSchemaVersion, user = this.user)
             request.index(INDEX_MANAGEMENT_INDEX)
                 .id(request.rollup.id)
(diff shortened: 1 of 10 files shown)


downsrob:
Looks like the test and build workflows are failing due to code style errors.
You can run `./gradlew detekt` to run this code style workflow locally. I think the ones you added are:
```
VarCouldBeVal - [targetIndexResolvedName] at /home/runner/work/index-management/index-management/src/main/kotlin/org/opensearch/indexmanagement/rollup/action/mapping/TransportUpdateRollupMappingAction.kt:63:9
VarCouldBeVal - [targetIndexResovledName] at /home/runner/work/index-management/index-management/src/main/kotlin/org/opensearch/indexmanagement/rollup/action/mapping/TransportUpdateRollupMappingAction.kt:118:9
VarCouldBeVal - [targetIndexResolvedName] at /home/runner/work/index-management/index-management/src/main/kotlin/org/opensearch/indexmanagement/rollup/RollupMapperService.kt:73:9
VarCouldBeVal - [targetIndexResolvedName] at /home/runner/work/index-management/index-management/src/main/kotlin/org/opensearch/indexmanagement/rollup/RollupIndexer.kt:127:13
```
Which shouldn't be hard to fix

downsrob:
We allow source index names to have wildcards, so we can have scenarios like this:
source_index: logs_etc_*
target_index: {{ctx.source_index}}_rollup
At index time the target index would then be resolved to logs_etc_*_rollup and then the indexing would fail because you would be trying to write to an index with a wildcard in the name. It is also a bit of an ambiguous customer experience, as a user might expect the target index in this case to be something like logs_etc_1_rollup, logs_etc_2_rollup, etc. based on the actual source index.

To prevent these failures at runtime and clarify the behavior let's validate that the user isn't using both wildcard source indices and a template in their target index. 

I think an easy way would just be to temporarily resolve the target index name in the init block of Rollup.kt and validate that it doesn't contain any wildcards. I don't mean updating the target index name with the resolved template, just resolving here to perform input validation. Feel free to implement it however you find best though!

src/main/kotlin/org/opensearch/indexmanagement/rollup/util/RollupFieldValueExpressionResolver.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.opensearch.indexmanagement.indexstatemanagement.util.XCONTENT_WITHOUT_TYPE
+import org.opensearch.indexmanagement.rollup.model.Rollup
+import org.opensearch.script.Script
+import org.opensearch.script.ScriptService
+import org.opensearch.script.ScriptType
+import org.opensearch.script.TemplateScript
+
+object RollupFieldValueExpressionResolver {
+
+    private val validTopContextFields = setOf("source_index")
```
downsrob, line 21: Could you change this to the Rollup.kt constant SOURCE_INDEX_FIELD

src/main/kotlin/org/opensearch/indexmanagement/rollup/util/RollupFieldValueExpressionResolver.kt
```diff
@@ -0,0 +1,46 @@
...
+    private lateinit var scriptService: ScriptService
+
+    fun resolve(rollup: Rollup, fieldValue: String): String {
+        val script = Script(ScriptType.INLINE, Script.DEFAULT_TEMPLATE_LANG, fieldValue, mapOf())
+
+        val contextMap = XContentHelper.convertToMap(
+            BytesReference.bytes(rollup.toXContent(XContentFactory.jsonBuilder(), XCONTENT_WITHOUT_TYPE)),
+            false,
+            XContentType.JSON
+        ).v2().filterKeys { key ->
```
downsrob, line 32: We have a helper function to get rid of some of this converting to map logic in OpenSearchExtensions.kt
Can be compressed to:
```
 val contextMap = rollup.toXContent(XContentFactory.jsonBuilder(), XCONTENT_WITHOUT_TYPE)
    .toMap()
    .filterKeys { it in validTopContextFields }
```

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/RollupActionIT.kt
```diff
@@ -146,6 +146,76 @@ class RollupActionIT : IndexStateManagementRestTestCase() {
...
+                ContentType.APPLICATION_JSON
+            )
+        )
+        client().makeRequest("PUT", "/_data_stream/$dataStreamName")
+
+        // Ensure rollup works on backing indices of a data stream.
+        val indexName = DataStream.getDefaultBackingIndexName(dataStreamName, 1L)
+        assertIndexRolledUp(indexName, policyID, rollup)
+
+        val catIndex = (cat("indices/rollup_$indexName?format=json&bytes=b") as List<Map<String, Any>>)
```
downsrob, line 213: Wow, I didn't realize we had this, super cool and useful to know, thank you.

That said, there is a helper function `assertIndexExists` which more succinctly checks this if you would like.

src/main/kotlin/org/opensearch/indexmanagement/rollup/action/index/TransportIndexRollupAction.kt
```diff
@@ -143,6 +144,14 @@ class TransportIndexRollupAction @Inject constructor(
...
 
         private fun putRollup() {
+            if (!validateTargetIndexName()) {
+                return actionListener.onFailure(
+                    OpenSearchStatusException(
+                        "target_index value is invalid: ${request.rollup.targetIndex}",
+                        RestStatus.BAD_REQUEST
+                    )
+                )
+            }
```
downsrob, line 154: I would prefer this block be moved to onGetRollup, just before putRollup
I just want to maintain that putRollup only worries about adding the rollup, and keep all validation in onGetRollup rather than letting it bleed between multiple functions

downsrob:
Thanks Petar, I am happy to approve after these.

downsrob (approved):

Angie-Zhang (approved):

PR #445, opened 2022-08-05
alias in  rollup target_index field

petardz:
Signed-off-by: Petar Dzepina <petar.dzepina@gmail.com>

*Issue #, if available:* #61 

*Description of changes:*

Added support for using alias in rollup's target_index field

*CheckList:*
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/index-management/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #575, opened 2022-10-20
Adds an alias action (#35)

goyamegh:
Signed-off-by: Megha Goyal <goyamegh@amazon.com>

*Issue #, if available:*
 #35 (#304)

*Description of changes:* 
Adds support for an alias action in ISM.
Gives complete support for all update alias API options by reusing the alias actions parser internally and explicitly not defining the mappings in our config index.

Coming from the GitHub feature request, the use cases people asked for:

> 1. My use case is, If we have an alias for an index with rollover ILM policy, can we remove it from alias after some predefined days, let's say 30 days? I know there is a way to delete an index once its old, but in our case we do not want to delete but want the index to be removed from alias and have the alias switch and point to indices whose age is less than 30 days.

 This should be easily doable with the Alias action, just have a state that gets transitioned into after the index is 30 days old (or 30 days rollover age) and then remove the alias.

> 2. My use case is: we provide search_alias to search our dataset. We publish new version of dataset every week as a new index (e.g. dataset-yyyy-MM-dd. We want the search_alias always pointing to the latest published dataset. The process is: ingest new data to a new index, once ingestion finish, re-pointing search_alias to the new index, remove old index from the search_alias

Should be possible, the newly created index can remove the alias from all search-index-* indices and then add the alias it itself in a single call. The main thing would be knowing when ingestion finished for the new index.. which you could just use some absolute time using the transition condition if you know it only takes x amount of time.

*CheckList:*

**Testing steps:**

1. Rollover index with deletion of indices when the doc count exceeds a certain value.
1.1. Create an ISM policy
```
curl -X PUT "localhost:9200/_plugins/_ism/policies/rollover_policy?pretty" -H 'Content-Type: application/json' -d'
{
  "policy": {
    "description": "Example rollover policy.",
    "default_state": "rollover",
    "states": [
      {
        "name": "rollover",
        "actions": [
          {
            "rollover": {
              "min_doc_count": 1
            }
          }
        ],
        "transitions": [{
            "state_name": "alias",
            "conditions": {
              "min_doc_count": "2"
            }
          }]
      },
      {
        "name": "alias",
        "actions": [
          {
            "alias": {
              "actions": [
                {
                  "remove": {
                      "alias": "log"
                  }
                }
              ]
            }
          }
        ]
      }
    ],
    "ism_template": {
      "index_patterns": ["log*"],
      "priority": 100
    }
  }
}
'
```
1.2 Create an index template to enable the policy on:
```
curl -X PUT "localhost:9200/_index_template/ism_rollover1?" -H 'Content-Type: application/json' -d'
{
  "index_patterns": ["1log*"],
  "template": {
   "settings": {
    "plugins.index_state_management.rollover_alias": "1log"
   }
 }
}
'
```
1.3 Change the cluster settings to trigger jobs every minute:
```
curl -X PUT "localhost:9200/_cluster/settings?pretty=true" -H 'Content-Type: application/json' -d'
{
  "persistent" : {
    "plugins.index_state_management.job_interval" : 1
  }
}
'
```
1.4 Create a new index:
```
curl -X PUT "localhost:9200/log-000001" -H 'Content-Type: application/json' -d'
{
  "aliases": {
    "log": {
      "is_write_index": true
    }
  }
}
'
```
1.5 Add a document to the index to trigger the job:
```

curl -X POST "localhost:9200/log-000001/_doc" -H 'Content-Type: application/json' -d'
{
  "message": "dummy"
}
'
```
1.6 The first job will trigger the rollover action and a new index will be created. 
1.7 Add another document to the two indices like step 1.5.
1.8 The new job will cause the second index to point to the log alias and the older one will be removed due to alias action.
1.9. Verify these steps using alias and index API:
```
curl -X GET "localhost:9200/_cat/indices?pretty"

curl -X GET "localhost:9200/_cat/aliases?pretty"
```

2. Fail the case if a user tries to remove index using alias action.
2.1 Create the following policy:
```
curl -X PUT "localhost:9200/_plugins/_ism/policies/remove_index_policy12345?pretty" -H 'Content-Type: application/json' -d'
{
  "policy": {
    "description": "Example rollover policy.",
    "default_state": "rollover",
    "states": [
      {
        "name": "rollover",
        "actions": [
          {
            "rollover": {
              "min_doc_count": 1
            }
          }
        ],
        "transitions": [{
            "state_name": "alias",
            "conditions": {
              "min_doc_count": "2"
            }
          }]
      },
      {
        "name": "alias",
        "actions": [
          {
            "alias": {
              "actions": [
                {
                  "add": {
                      "alias": "log"
                  }
                },
                {
                  "remove": {
                      "index": "log-00001"
                  }
                }
              ]
            }
          }
        ]
      }
    ],
    "ism_template": {
      "index_patterns": ["MMMMMMMMttestlog*"],
      "priority": 100
    }
  }
}
'
```
2.2 This will create an Illegal argument exception with the error: "Remove_index is not allowed here."

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasAction.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasAction.kt
@@ -1,0 +1,51 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.indexmanagement.indexstatemanagement.action
+
+import org.opensearch.action.admin.indices.alias.IndicesAliasesRequest
+import org.opensearch.common.io.stream.StreamOutput
+import org.opensearch.common.xcontent.ToXContent
+import org.opensearch.common.xcontent.XContentBuilder
+import org.opensearch.indexmanagement.indexstatemanagement.step.alias.AttemptAliasStep
+import org.opensearch.indexmanagement.spi.indexstatemanagement.Action
+import org.opensearch.indexmanagement.spi.indexstatemanagement.Step
+import org.opensearch.indexmanagement.spi.indexstatemanagement.model.StepContext
+
+class AliasAction(
+    val actions: List<IndicesAliasesRequest.AliasActions>,
+    index: Int
+) : Action(name, index) {
+
+    init {
+        require(actions.isNotEmpty()) { "At least one alias action needs to be specified." }
+    }
+
+    private val attemptAliasStep = AttemptAliasStep(this)
+
+    private val steps = listOf(attemptAliasStep)
+
+    override fun getStepToExecute(context: StepContext): Step {
+        return attemptAliasStep
+    }
+
+    override fun getSteps(): List<Step> = steps
+
+    override fun populateAction(builder: XContentBuilder, params: ToXContent.Params) {
+        builder.startObject(type)
+        builder.field(ACTIONS, actions)
+        builder.endObject()
+    }
+
+    override fun populateAction(out: StreamOutput) {
+        out.writeList(actions)
+        out.writeInt(actionIndex)
+    }
+
+    companion object {
+        const val name = "alias"
+        const val ACTIONS = "actions"
+    }
+}
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasAction.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasAction.kt
@@ -9,7 +9,7 @@ import org.opensearch.action.admin.indices.alias.IndicesAliasesRequest
 import org.opensearch.common.io.stream.StreamOutput
 import org.opensearch.common.xcontent.ToXContent
 import org.opensearch.common.xcontent.XContentBuilder
-import org.opensearch.indexmanagement.indexstatemanagement.step.alias.AttemptAliasStep
+import org.opensearch.indexmanagement.indexstatemanagement.step.alias.AttemptAliasActionsStep
 import org.opensearch.indexmanagement.spi.indexstatemanagement.Action
 import org.opensearch.indexmanagement.spi.indexstatemanagement.Step
 import org.opensearch.indexmanagement.spi.indexstatemanagement.model.StepContext
@@ -19,18 +19,28 @@ class AliasAction(
     index: Int
 ) : Action(name, index) {
 
+    /**
+     * Allowing the alias action to be only applicable on the managed index for ADD and REMOVE actions only.
+     * https://github.com/opensearch-project/OpenSearch/blob/4d045a164e12a382881140e32f9285a3224fecc7/server/src/main/java/org/opensearch/action/admin/indices/alias/IndicesAliasesRequest.java#L105
+     */
     init {
         require(actions.isNotEmpty()) { "At least one alias action needs to be specified." }
-        require(!actions.any { it.actionType() == IndicesAliasesRequest.AliasActions.Type.REMOVE_INDEX }) { "Remove_index is not allowed here." }
-        require(actions.all { it.indices().isNullOrEmpty() }) { "Alias actions are only allowed on managed indices." }
+        val allowedActionTypes = listOf(IndicesAliasesRequest.AliasActions.Type.ADD, IndicesAliasesRequest.AliasActions.Type.REMOVE)
+        require(actions.all { it.actionType() in allowedActionTypes }) { "Only ADD and REMOVE actions are allowed." }
+        require(
+            actions.all { it.indices().isNullOrEmpty() }
+        ) { "Alias action can only work on its applied index so don't accept index/indices parameter." }
+        require(
+            actions.all { it.aliases().isNotEmpty() }
+        ) { "At least one alias needs to be specified." }
     }
 
-    private val attemptAliasStep = AttemptAliasStep(this)
+    private val attemptAliasActionsStep = AttemptAliasActionsStep(this)
 
-    private val steps = listOf(attemptAliasStep)
+    private val steps = listOf(attemptAliasActionsStep)
 
     override fun getStepToExecute(context: StepContext): Step {
-        return attemptAliasStep
+        return attemptAliasActionsStep
     }
 
     override fun getSteps(): List<Step> = steps
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasAction.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasAction.kt
@@ -21,6 +21,8 @@ class AliasAction(
 
     init {
         require(actions.isNotEmpty()) { "At least one alias action needs to be specified." }
+        require(!actions.any { it.actionType() == IndicesAliasesRequest.AliasActions.Type.REMOVE_INDEX }) { "Remove_index is not allowed here." }
+        require(actions.all { it.indices().isNullOrEmpty() }) { "Alias actions are only allowed on managed indices." }
     }
 
     private val attemptAliasStep = AttemptAliasStep(this)
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasActionParser.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasActionParser.kt
@@ -1,0 +1,45 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.indexmanagement.indexstatemanagement.action
+
+import org.opensearch.action.admin.indices.alias.IndicesAliasesRequest
+import org.opensearch.common.io.stream.StreamInput
+import org.opensearch.common.xcontent.XContentParser
+import org.opensearch.common.xcontent.XContentParser.Token
+import org.opensearch.common.xcontent.XContentParserUtils.ensureExpectedToken
+import org.opensearch.indexmanagement.indexstatemanagement.action.AliasAction.Companion.ACTIONS
+import org.opensearch.indexmanagement.spi.indexstatemanagement.Action
+import org.opensearch.indexmanagement.spi.indexstatemanagement.ActionParser
+
+class AliasActionParser : ActionParser() {
+    override fun fromStreamInput(sin: StreamInput): Action {
+        val actions = sin.readList(IndicesAliasesRequest::AliasActions)
+        val index = sin.readInt()
+        return AliasAction(actions, index)
+    }
+
+    override fun fromXContent(xcp: XContentParser, index: Int): Action {
+        val actions: MutableList<IndicesAliasesRequest.AliasActions> = mutableListOf()
+
+        ensureExpectedToken(Token.START_OBJECT, xcp.currentToken(), xcp)
+        while (xcp.nextToken() != Token.END_OBJECT) {
+            val fieldName = xcp.currentName()
+            xcp.nextToken()
+            when (fieldName) {
+                ACTIONS -> {
+                    ensureExpectedToken(Token.START_ARRAY, xcp.currentToken(), xcp)
+                    while (xcp.nextToken() != Token.END_ARRAY) {
+                        actions.add(IndicesAliasesRequest.AliasActions.fromXContent(xcp))
+                    }
+                }
+                else -> throw IllegalArgumentException("Invalid field: [$fieldName] found in AliasAction.")
+            }
+        }
+        return AliasAction(actions, index)
+    }
+
+    override fun getActionType(): String = AliasAction.name
+}
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasActionParser.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasActionParser.kt
@@ -5,6 +5,7 @@
 
 package org.opensearch.indexmanagement.indexstatemanagement.action
 
+import org.apache.logging.log4j.LogManager
 import org.opensearch.action.admin.indices.alias.IndicesAliasesRequest
 import org.opensearch.common.io.stream.StreamInput
 import org.opensearch.common.xcontent.XContentParser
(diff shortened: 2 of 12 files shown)


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasAction.kt
```diff
@@ -0,0 +1,51 @@
...
+import org.opensearch.common.io.stream.StreamOutput
+import org.opensearch.common.xcontent.ToXContent
+import org.opensearch.common.xcontent.XContentBuilder
+import org.opensearch.indexmanagement.indexstatemanagement.step.alias.AttemptAliasStep
+import org.opensearch.indexmanagement.spi.indexstatemanagement.Action
+import org.opensearch.indexmanagement.spi.indexstatemanagement.Step
+import org.opensearch.indexmanagement.spi.indexstatemanagement.model.StepContext
+
+class AliasAction(
+    val actions: List<IndicesAliasesRequest.AliasActions>,
```
khushbr, line 18: I am trying to understand why this is a list ?

[AliasActions](https://github.com/opensearch-project/OpenSearch/blob/fe3994ccdf62735907bc4834f93c71d97636d150/server/src/main/java/org/opensearch/action/admin/indices/alias/IndicesAliasesRequest.java#L101) by itself is defined as ` Request to take one or more actions on one or more indexes and alias combinations.`

khushbr: See https://github.com/opensearch-project/OpenSearch/blob/4d045a164e12a382881140e32f9285a3224fecc7/server/src/main/java/org/opensearch/action/admin/indices/alias/IndicesAliasesRequest.java#L105

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasActionParser.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    override fun fromXContent(xcp: XContentParser, index: Int): Action {
+        val actions: MutableList<IndicesAliasesRequest.AliasActions> = mutableListOf()
+
+        ensureExpectedToken(Token.START_OBJECT, xcp.currentToken(), xcp)
+        while (xcp.nextToken() != Token.END_OBJECT) {
+            val fieldName = xcp.currentName()
+            xcp.nextToken()
+            when (fieldName) {
+                ACTIONS -> {
```
bowenlan-amzn, line 32: Block `remove_index` action
Block `index` `indices` fields if user provided

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/alias/AttemptAliasStep.kt
```diff
@@ -0,0 +1,103 @@
...
+                // Take the indices that were set to be updated on the alias action and join them into a single string
+                // So we don't have to pass each one individually into the script service to be compiled
+                val indices = it.indices().joinToString(",")
+                // Compile them so the user can use the current dynamic index name in the name such as "{{ctx.index}}" in the static policy
+                val indicesScript = Script(ScriptType.INLINE, Script.DEFAULT_TEMPLATE_LANG, indices, mapOf())
+                val compiledIndices = compileTemplate(indicesScript, context.metadata, indices, context.scriptService)
+                // Split them back into individual index names and set back on the AliasActions request
+                val splitIndices = Strings.splitStringByCommaToArray(compiledIndices)
+                it.indices(*splitIndices)
+                request.addAliasAction(it)
```
bowenlan-amzn, line 44: Populate the `index` to be the managed index itself

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasActionParser.kt
goyamegh: Added checks in initiate function to raise Illegal argument exception.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/alias/AttemptAliasStep.kt
goyamegh: Removed the logic for user input and using the managed index itself.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasAction.kt
```diff
@@ -21,6 +21,8 @@ class AliasAction(
 
     init {
         require(actions.isNotEmpty()) { "At least one alias action needs to be specified." }
+        require(!actions.any { it.actionType() == IndicesAliasesRequest.AliasActions.Type.REMOVE_INDEX }) { "Remove_index is not allowed here." }
+        require(actions.all { it.indices().isNullOrEmpty() }) { "Alias actions are only allowed on managed indices." }
```
bowenlan-amzn, line 25: need to verify both indices and index to be empty.
And change the message to be "Alias action can only work on its applied index so don't accept index/indices parameter."

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/alias/AttemptAliasStep.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.logging.log4j.LogManager
+import org.opensearch.action.admin.indices.alias.IndicesAliasesRequest
+import org.opensearch.action.support.master.AcknowledgedResponse
+import org.opensearch.indexmanagement.indexstatemanagement.action.AliasAction
+import org.opensearch.indexmanagement.opensearchapi.suspendUntil
+import org.opensearch.indexmanagement.spi.indexstatemanagement.Step
+import org.opensearch.indexmanagement.spi.indexstatemanagement.model.ManagedIndexMetaData
+import org.opensearch.indexmanagement.spi.indexstatemanagement.model.StepMetaData
+
+class AttemptAliasStep(private val action: AliasAction) : Step(name) {
```
khushbr, line 17: We are attempting to update multiple aliases for the index, let us rename this.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/alias/AttemptAliasStep.kt
```diff
@@ -0,0 +1,77 @@
...
+            val response: AcknowledgedResponse = context.client.admin().indices().suspendUntil { aliases(request, it) }
+            handleResponse(response, indexName)
+        } catch (e: Exception) {
+            handleException(e, indexName)
+        }
+        return this
+    }
+
+    private fun handleException(e: Exception, indexName: String) {
+        val message = getFailedMessage(indexName)
```
khushbr, line 42: Add the aliases information to this error message.

khushbr:
The `Rollup alias` work is separately under review at https://github.com/opensearch-project/index-management/pull/445/files, so let us exclude `Rollup` action from this Alias code-path.  


The question from @bowenlan-amzn has still not been addressed `We can see there're use cases some actions (shrink, rollover, rollup) create new index and we would like some alias to be add to the new index.`. 

Most often, the alias action will work in conjunction with another action (As multiple use-cases have also pointed), and in cases the new index is created, the ISM workflow must provide in-built support to add alias-ing to the new index (some options were laid out in Bowen's comment).

bowenlan-amzn:
> The Rollup alias work is separately under review at https://github.com/opensearch-project/index-management/pull/445/files, so let us exclude Rollup action from this Alias code-path.

If an index is under a rollup target alias, ISM policy should not remove this alias from it. Is this what you are thinking about? We can add an issue to enhance this later. @khushbr

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasAction.kt
goyamegh: 1. This is not needed as we store them as a string list in the indices field only. Refer this line: https://github.com/opensearch-project/OpenSearch/blob/4d045a164e12a382881140e32f9285a3224fecc7/server/src/main/java/org/opensearch/action/admin/indices/alias/IndicesAliasesRequest.java#L360
2. Changed the message.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasAction.kt
```diff
@@ -0,0 +1,53 @@
...
+import org.opensearch.indexmanagement.spi.indexstatemanagement.model.StepContext
+
+class AliasAction(
+    val actions: List<IndicesAliasesRequest.AliasActions>,
+    index: Int
+) : Action(name, index) {
+
+    init {
+        require(actions.isNotEmpty()) { "At least one alias action needs to be specified." }
+        require(!actions.any { it.actionType() == IndicesAliasesRequest.AliasActions.Type.REMOVE_INDEX }) { "Remove_index is not allowed here." }
```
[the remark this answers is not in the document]
goyamegh: 1. Added the check.
2. Made changes on action type to check for ADD and REMOVE.
3. Added the KDoc.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/alias/AttemptAliasStep.kt
goyamegh: Done.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/alias/AttemptAliasStep.kt
goyamegh: Renamed to AttemptAliasActionsStep

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/alias/AttemptAliasStep.kt
```diff
@@ -0,0 +1,77 @@
...
+        logger.error(message, e)
+        stepStatus = StepStatus.FAILED
+        val mutableInfo = mutableMapOf("message" to message)
+        val errorMessage = e.message
+        if (errorMessage != null) mutableInfo["cause"] = errorMessage
+        info = mutableInfo.toMap()
+    }
```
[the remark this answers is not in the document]
goyamegh: Found one more file, but if we extract it, then we need to return info and step status as new objects, as they are getting modified in the function. Keeping it the same for now, but will figure out a better way to handle this and raise another PR.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/alias/AttemptAliasStep.kt
```diff
@@ -0,0 +1,77 @@
...
+import org.opensearch.indexmanagement.spi.indexstatemanagement.model.StepMetaData
+
+class AttemptAliasStep(private val action: AliasAction) : Step(name) {
+
+    private val logger = LogManager.getLogger(javaClass)
+    private var stepStatus = StepStatus.STARTING
+    private var info: Map<String, Any>? = null
+
+    override suspend fun execute(): Step {
+        val context = this.context ?: return this
```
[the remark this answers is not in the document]
goyamegh: Looks like it is being used in the same manner in all parsers. Kindly refer: https://github.com/opensearch-project/index-management/blob/2d3fd977168063f1e8ecc29018e897e629352392/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/allocation/AttemptAllocationStep.kt#L25, https://github.com/opensearch-project/index-management/blob/efe2d332ad66db308fd172e5f3cae4777cece9b9/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/close/AttemptCloseStep.kt#L26

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasActionParser.kt
```diff
@@ -0,0 +1,45 @@
...
+                    ensureExpectedToken(Token.START_ARRAY, xcp.currentToken(), xcp)
+                    while (xcp.nextToken() != Token.END_ARRAY) {
+                        actions.add(IndicesAliasesRequest.AliasActions.fromXContent(xcp))
+                    }
+                }
+                else -> throw IllegalArgumentException("Invalid field: [$fieldName] found in AliasAction.")
```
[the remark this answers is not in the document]
goyamegh: Added the log.

khushbr:
> > The Rollup alias work is separately under review at https://github.com/opensearch-project/index-management/pull/445/files, so let us exclude Rollup action from this Alias code-path.
> 
> If an index is under a rollup target alias, ISM policy should not remove this alias from it. Is this what you are thinking about? We can add an issue to enhance this later. @khushbr

That is fine by me. @goyamegh Let us open an issue to track the `conflicting alias action not allowed.`

khushbr (approved):

bowenlan-amzn (approved):

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AliasActionIT.kt
```diff
@@ -0,0 +1,150 @@
...
+
+        waitFor {
+            val info = getExplainManagedIndexMetaData(indexName).info as Map<String, Any?>
+            assertEquals("Alias was not successfully updated", AttemptAliasActionsStep.getSuccessMessage(indexName), info["message"])
+            val alias = getAlias(indexName, "")
+            assertTrue("Alias was not added to index", alias.containsKey(aliasName))
+        }
+    }
+
+    fun `test adding alias to index using ctx`() {
```
bowenlan-amzn, line 65: What's different between this one and the first test `test adding alias to index`?

goyamegh: This was added earlier when the changes were intended for any index, not just managed index. As discussed, feel free to remove this as it is redundant now.

madhukarmmallia-plivo:
@goyamegh @bowenlan-amzn : I have a query here with respect to dynamic aliases. If the policy is used for multiple index name patterns, can this feature be used for dynamically adding an alias using the mustache template (for example by passing `{{ ctx.index }}` to the `add` step in the `alias`?

bowenlan-amzn:
> @goyamegh @bowenlan-amzn : I have a query here with respect to dynamic aliases. If the policy is used for multiple index name patterns, can this feature be used for dynamically adding an alias using the mustache template (for example by passing `{{ ctx.index }}` to the `add` step in the `alias`?

@madhukarmmallia-plivo
I am not sure `ctx.index` as the alias name is useful, because this alias is gonna be different for every indices, but normally one alias contains a group of indices. Maybe what you want is to parse sth from the index name to be used as the alias? I suggest to create different policy for different index patterns, and in each one you can define a different alias name to add.
Feel free to open a feature request issue to discuss more on this.

PR #652, opened 2023-01-06
Add option for bwc tests in bundle level

zelinh:
### Description
Add option to run BWC tests in distribution level, which is running tests in the test cluster with latest distribution bundle installed and all plugins (included in the latest manifest) present. 

The command to run BWC tests in distribution level would be `./gradlew bwcTestSuite -Dtests.security.manager=false -PcustomDistributionDownloadType=bundle`
If property `customDistributionDownloadType` is not set nor set to `bundle`, the BWC tests would be default to run in plugins level as previously configured. 

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-build/issues/2870

*CheckList:*


build.gradle
```diff
@@ -627,8 +627,13 @@ task integTestRemote(type: RestIntegTestTask)  {
     exclude 'org/opensearch/indexmanagement/indexstatemanagement/action/SnapshotActionIT.class'
 }
 
-String bwcVersion = "1.13.2.0"
+String bwcMinVersion = "1.13.2.0"
 String bwcJobSchedulerVersion = "1.13.0.0"
+String bwcBundleVersion = "1.3.2.0"
+Boolean bwcBundleTest = (project.findProperty('customDistributionDownloadType') != null &&
```
bowenlan-amzn, line 633: Just a bwcBundle property should be good enough, don't see why use `customDistributionDownloadType`

build.gradle
```diff
@@ -627,8 +627,13 @@ task integTestRemote(type: RestIntegTestTask)  {
     exclude 'org/opensearch/indexmanagement/indexstatemanagement/action/SnapshotActionIT.class'
 }
 
-String bwcVersion = "1.13.2.0"
+String bwcMinVersion = "1.13.2.0"
```
bowenlan-amzn, line 630: bwcPluginVersion may be better

build.gradle
```diff
@@ -627,8 +627,13 @@ task integTestRemote(type: RestIntegTestTask)  {
...
 String bwcJobSchedulerVersion = "1.13.0.0"
+String bwcBundleVersion = "1.3.2.0"
+Boolean bwcBundleTest = (project.findProperty('customDistributionDownloadType') != null &&
+        project.properties['customDistributionDownloadType'] == "bundle");
+String bwcVersion = bwcBundleTest ? bwcBundleVersion : bwcMinVersion
+String bwcCurrentVersion = opensearch_version.replace("-SNAPSHOT", "")
```
bowenlan-amzn, line 636: Why not use the existing variable opensearch_version?

build.gradle
```diff
@@ -641,38 +646,67 @@ getPluginResource(bwc_js_resource_location, bwc_js_download_url)
...
             numberOfNodes = 3
-            plugin(provider(new Callable<RegularFile>(){
-                @Override
-                RegularFile call() throws Exception {
-                    return new RegularFile() {
-                        @Override
-                        File getAsFile() {
-                            return fileTree(bwc_js_resource_location).getSingleFile()
+            if (bwcBundleTest) {
+                versions = ["1.3.2", bwcCurrentVersion]
```
bowenlan-amzn, line 658: get this 1.3.2 from `bwcBundleVersion`?

build.gradle
```diff
@@ -641,38 +646,67 @@ getPluginResource(bwc_js_resource_location, bwc_js_download_url)
...
+                    node.setting("plugins.security.ssl.http.pemtrustedcas_filepath", "root-ca.pem")
+                    node.setting("plugins.security.allow_unsafe_democertificates", "true")
+                    node.setting("plugins.security.allow_default_init_securityindex", "true")
+                    node.setting("plugins.security.authcz.admin_dn", "CN=kirk,OU=client,O=client,L=test,C=de")
+                    node.setting("plugins.security.audit.type", "internal_elasticsearch")
+                    node.setting("plugins.security.enable_snapshot_restore_privilege", "true")
+                    node.setting("plugins.security.check_snapshot_restore_write_privileges", "true")
+                    node.setting("plugins.security.restapi.roles_enabled", "[\"all_access\", \"security_rest_api_access\"]")
+                    node.setting("plugins.security.system_indices.enabled", "true")
+                }
```
bowenlan-amzn, line 682: See if you can move this to where you set up the test cluster using bundle

build.gradle
zelinh: The way we initiate the BWC tests running against bundle test cluster is to specify this gradle property. 
e.g. command would be like `./gradlew bwcTestSuite -Dtests.security.manager=false -PcustomDistributionDownloadType=bundle`
This property name is used across repo and set in the framework from core repo.

build.gradle
zelinh: I think bwcPluginVersion might be confusing as this version would be used for the test cluster set up with just opensearch min artifact. I would like to change if you think it's necessary.

build.gradle
zelinh: For the bwc tests on plugin level, we are getting the min artifacts from URL: `https://artifacts.opensearch.org/releases/core/opensearch/1.2.0/opensearch-min-1.2.0-linux-x64.tar.gz` for released version and `https://artifacts.opensearch.org/snapshots/core/opensearch/2.5.0-SNAPSHOT/opensearch-min-2.5.0-SNAPSHOT-linux-x64-latest.tar.gz` for unreleased snapshot version;
while we are getting the bundle from CI bucket with URL as `https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/2.5.0/latest/linux/x64/tar/dist/opensearch/opensearch-2.5.0-linux-x64.tar.gz` which doesn't contain `-SNAPSHOT` in the path.

build.gradle
zelinh: `bwcBundleVersion` is actually 4 digits version and this would need to be 3 digits.
Since this would be fixed most of the time as the old version, we don't need to change it every time. We could add another line to derive 3 digits version from `bwcBundleVersion` if you think that would be better.

build.gradle
zelinh: This won't be applicable at this time because as far as i see we want to make the framework itself more generic. This codebase is actually derived from security repo on running bwc tests. It works as parameter to add certificates to the test cluster.

build.gradle
bowenlan-amzn: Wondering if there's a guide line of the bwc version, for example, on 2.5 branch, should we run bwc against the the latest 1.x version or the last 2.4.x version.
If it's 1.x version, then true we don't change this often I suppose.

build.gradle
bowenlan-amzn: Thanks. I guess this effort is to enable the bwc bundle test as github workflow. Not sure if the latest of CI bucket is stable enough for running workflow 😛

build.gradle
bowenlan-amzn: I see. I though this is only used here for this variable.

build.gradle
bowenlan-amzn: Oh I remember sth on this. The bundle would always include security plugin, but this attribute can only be set here, the script to set this attribute from security plugin cannot work with the test framework.

build.gradle
```diff
@@ -641,38 +646,67 @@ getPluginResource(bwc_js_resource_location, bwc_js_download_url)
...
+                    node.setting("plugins.security.ssl.transport.pemkey_filepath", "esnode-key.pem")
+                    node.setting("plugins.security.ssl.transport.pemtrustedcas_filepath", "root-ca.pem")
+                    node.setting("plugins.security.ssl.transport.enforce_hostname_verification", "false")
+                    node.setting("plugins.security.ssl.http.enabled", "true")
+                    node.setting("plugins.security.ssl.http.pemcert_filepath", "esnode.pem")
+                    node.setting("plugins.security.ssl.http.pemkey_filepath", "esnode-key.pem")
+                    node.setting("plugins.security.ssl.http.pemtrustedcas_filepath", "root-ca.pem")
+                    node.setting("plugins.security.allow_unsafe_democertificates", "true")
+                    node.setting("plugins.security.allow_default_init_securityindex", "true")
+                    node.setting("plugins.security.authcz.admin_dn", "CN=kirk,OU=client,O=client,L=test,C=de")
```
bowenlan-amzn, line 676: Change this to `node.setting("plugins.security.authcz.admin_dn", "\n - CN=kirk,OU=client,O=client,L=test,C=de")`

I worked on this not long ago. This attribute is a yml list, so have to do `\n - ` before the attribute, otherwise this attribute would be ignored.

build.gradle
bowenlan-amzn: Yeah, our standpoint is different 😛 you are from infra, I am looking this from plugin. This variable will be used to set bwc plugin version, and also min version haha. I am ok to keep this

bowenlan-amzn (approved):

AWSHurneyt (approved):

Issue #643, opened 2023-01-05
[FEATURE] Close action validation


Angie-Zhang:
Clarify Close action validation conditions and implement it.
 
**Meta Feature**
https://github.com/opensearch-project/index-management/issues/587

**Validation Conditions**
1. Error Prevention checks if the index exists and its name is valid. If not, suspend the close action and set ValidationStatus = RE_VALIDATING.
2. Indices have snapshot in progress can't be closed. 

To resolve #27 , we need a policy validator to prevent invalid actions on closed indies.

PR #728, opened 2023-04-06
Error Prevention: Add close action

Angie-Zhang:
*Issue #, if available:*

https://github.com/opensearch-project/index-management/issues/643

*Description of changes:*
Add close action validation in Error Prevention framework

*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateClose.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateClose.kt
@@ -55,6 +55,7 @@ class ValidateClose(
             val message = getIndexNotValidMessage(indexName)
             logger.warn(message)
             validationMessage = message
+            return false
         }
         return true
     }
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateClose.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateClose.kt
@@ -63,8 +63,8 @@ class ValidateClose(
     @Suppress("TooManyFunctions")
     companion object {
         const val name = "validate_close"
-        fun getNoIndexMessage(index: String) = "no such index [index=$index]"
-        fun getIndexNotValidMessage(index: String) = "Close index [index=$index] not valid"
+        fun getNoIndexMessage(index: String) = "No such index [index=$index] for close action."
+        fun getIndexNotValidMessage(index: String) = "Index [index=$index] is not valid. Abort close action on it."
         fun getValidationPassedMessage(index: String) = "Close action validation passed for [index=$index]"
     }
 }
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateClose.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateClose.kt
@@ -27,6 +27,7 @@ class ValidateClose(
     override fun execute(indexName: String): Validate {
         // if these conditions are false, fail validation and do not execute close action
         if (!indexExists(indexName) || !validIndex(indexName)) {
+            validationStatus = ValidationStatus.FAILED
             return this
         }
         validationMessage = getValidationPassedMessage(indexName)
@@ -38,7 +39,6 @@ class ValidateClose(
         if (!isIndexExists) {
             val message = getNoIndexMessage(indexName)
             logger.warn(message)
-            validationStatus = ValidationStatus.RE_VALIDATING
             validationMessage = message
             return false
         }
@@ -54,7 +54,6 @@ class ValidateClose(
         } catch (e: Exception) {
             val message = getIndexNotValidMessage(indexName)
             logger.warn(message)
-            validationStatus = ValidationStatus.RE_VALIDATING
             validationMessage = message
         }
         return true
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateClose.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateClose.kt
@@ -1,0 +1,70 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.indexmanagement.indexstatemanagement.validation
+
+import org.apache.logging.log4j.LogManager
+import org.opensearch.cluster.metadata.MetadataCreateIndexService
+import org.opensearch.cluster.service.ClusterService
+import org.opensearch.common.settings.Settings
+import org.opensearch.indexmanagement.spi.indexstatemanagement.Validate
+import org.opensearch.indexmanagement.util.OpenForTesting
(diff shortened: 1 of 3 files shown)


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateClose.kt
```diff
@@ -0,0 +1,70 @@
...
+        val exceptionGenerator: (String, String) -> RuntimeException = { index_name, reason -> InvalidIndexNameException(index_name, reason) }
+        // If the index name is invalid for any reason, this will throw an exception giving the reason why in the message.
+        // That will be displayed to the user as the cause.
+        try {
+            MetadataCreateIndexService.validateIndexOrAliasName(indexName, exceptionGenerator)
+        } catch (e: Exception) {
+            val message = getIndexNotValidMessage(indexName)
+            logger.warn(message)
+            validationStatus = ValidationStatus.RE_VALIDATING
+            validationMessage = message
```
bowenlan-amzn, line 58: Do we need to return false here?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateClose.kt
```diff
@@ -0,0 +1,70 @@
...
+        validationMessage = getValidationPassedMessage(indexName)
+        return this
+    }
+
+    private fun indexExists(indexName: String): Boolean {
+        val isIndexExists = clusterService.state().metadata.indices.containsKey(indexName)
+        if (!isIndexExists) {
+            val message = getNoIndexMessage(indexName)
+            logger.warn(message)
+            validationStatus = ValidationStatus.RE_VALIDATING
```
bowenlan-amzn, line 41: If index doesn't exist, I think it's better to be in the Failed status

When we check for index existence, cold tier index doesn't exist in cluster state anymore, will need to use another way to retrieve, check this part to understand https://github.com/opensearch-project/index-management/blob/main/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt#L268-L275

bowenlan-amzn:
A general question: if validation failed, can user retry the validation after fix the issue?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateClose.kt
Angie-Zhang: I think index type (i.e. warm/cold) only matters in AWS managed services. Do we need logic to handle them in open source side?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateClose.kt
```diff
@@ -0,0 +1,70 @@
...
+        try {
+            MetadataCreateIndexService.validateIndexOrAliasName(indexName, exceptionGenerator)
+        } catch (e: Exception) {
+            val message = getIndexNotValidMessage(indexName)
+            logger.warn(message)
+            validationStatus = ValidationStatus.RE_VALIDATING
+            validationMessage = message
+        }
+        return true
+    }
```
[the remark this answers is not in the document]
Angie-Zhang: To make reviewing and testing easier, there will be a individual PR for all refactoring changes, including moving reusable methods & logs to abstract level.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateClose.kt
bowenlan-amzn: Yes, we need to consider that. The validation logic may not be same for different index types, and you can see the link I provide how to deal with other index type using the interface.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateClose.kt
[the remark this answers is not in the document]
bowenlan-amzn: Ok, sounds good

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateClose.kt
Angie-Zhang: This is a good point. I think it is better to set a common method/logic to check index type for all actions. Let's do it in refactoring PR instead of here in one action's PR.

bowenlan-amzn: This would not be trivial, can add another TODO item in meta issue to track the refactor for the index type extension.

Angie-Zhang:
> A general question: if validation failed, can user retry the validation after fix the issue?

In ManagedIndexRunner, when the validation is enabled and the result is failed, it will call disableManagedIndexConfig method, which means the job is disabled. I think we should change the logic as the action failed, so user can call retry API for this action after they fix the issue.  This change can be part of the refactoring PR.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateClose.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    @Suppress("TooManyFunctions")
+    companion object {
+        const val name = "validate_close"
+        fun getNoIndexMessage(index: String) = "No such index [index=$index] for close action."
+        fun getIndexNotValidMessage(index: String) = "Index [index=$index] is not valid. Abort close action on it."
```
AWSHurneyt, line 66: Are we able to make this message more actionable by the users? For example, if the index name is invalid because it contains illegal characters, could the error message include a list of legal or illegal characters (whichever is less cumbersome to display)?

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateCloseIT.kt
```diff
@@ -0,0 +1,66 @@
...
+import org.opensearch.indexmanagement.spi.indexstatemanagement.Validate
+import org.opensearch.indexmanagement.waitFor
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+import java.util.Locale
+
+class ValidateCloseIT : IndexStateManagementRestTestCase() {
+    private val testIndexName = javaClass.simpleName.lowercase(Locale.ROOT)
+
+    fun `test basic`() {
```
AWSHurneyt, line 22: Nitpick: Could this test name be a little more descriptive? That could help future devs debug test failures.

bowenlan-amzn (approved):

bowenlan-amzn:
> > A general question: if validation failed, can user retry the validation after fix the issue?
> 
> In ManagedIndexRunner, when the validation is enabled and the result is failed, it will call disableManagedIndexConfig method, which means the job is disabled. I think we should change the logic as the action failed, so user can call retry API for this action after they fix the issue. This change can be part of the refactoring PR.

Got it, this is important before release, think of user cannot retry after validation failed is very bad... So not a refactor PR, a separate PR with an IT to cover this path is needed

AWSHurneyt (approved):

Issue #774, opened 2023-05-10
[BUG] ISM add policy API recursive permission check causes stackoverflow


bowenlan-amzn:
Code: https://github.com/opensearch-project/index-management/blob/main/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt#L158-L182

If user add policy to a large amount of indices in one call, this code logic just recursively call itself to check managed index permission one by one.
In the end, it will cause stackoverflow error and causes the node to go down.

```
[2023-05-09T14:38:56,649][ERROR][o.o.s.f.SecurityFilter   ] [3fd290699e29bc6eedad27f6fa967a48] Unexpected exception java.lang.StackOverflowError
java.lang.StackOverflowError
        at org.apache.logging.log4j.util.PropertiesUtil$Environment.access$200(PropertiesUtil.java:420)
        at org.apache.logging.log4j.util.PropertiesUtil.getStringProperty(PropertiesUtil.java:368)
        at org.apache.logging.log4j.util.PropertiesUtil.getBooleanProperty(PropertiesUtil.java:155)
        at org.apache.logging.log4j.status.StatusLogger.isDebugPropertyEnabled(StatusLogger.java:120)
        at org.apache.logging.log4j.status.StatusLogger.isEnabled(StatusLogger.java:403)
        at org.apache.logging.log4j.status.StatusLogger.isEnabled(StatusLogger.java:344)
        at org.apache.logging.log4j.spi.AbstractLogger.logIfEnabled(AbstractLogger.java:1930)
        at org.apache.logging.log4j.spi.AbstractLogger.warn(AbstractLogger.java:2805)
        at org.apache.logging.log4j.spi.AbstractLogger.handleLogMessageException(AbstractLogger.java:2225)
        at org.apache.logging.log4j.spi.AbstractLogger.tryLogMessage(AbstractLogger.java:2208)
        at org.apache.logging.log4j.spi.AbstractLogger.logMessageTrackRecursion(AbstractLogger.java:2159)
        at org.apache.logging.log4j.spi.AbstractLogger.logMessageSafely(AbstractLogger.java:2142)
        at org.apache.logging.log4j.spi.AbstractLogger.logMessage(AbstractLogger.java:2058)
        at org.apache.logging.log4j.spi.AbstractLogger.logIfEnabled(AbstractLogger.java:1931)
[...]

PR #779, opened 2023-05-15
AddPolicyAction - Recursion removal

petardz:
*Issue #, if available:* #774

*Description of changes:*

Removed recursion when checking permissions on indices.

Fixed index pattern permission check: injected "injected_user" property into threadContext 

*CheckList:*


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
```diff
@@ -145,55 +148,36 @@ class TransportAddPolicyAction @Inject constructor(
...
-                // sanity check that there are indices - if none then return
-                if (permittedIndices.isEmpty()) {
-                    actionListener.onResponse(ISMStatusResponse(0, failedIndices))
-                    return
-                }
-                // Filter out the indices that the user does not have permissions for
-                indicesToAdd.values.removeIf { !permittedIndices.contains(it) }
-                removeClosedIndices()
+            // sanity check that there are indices - if none then return
+            if (permittedIndices.isEmpty()) {
```
bowenlan-amzn, line 175: Change permittedIndices to a local variable

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
```diff
@@ -145,55 +148,36 @@ class TransportAddPolicyAction @Inject constructor(
                     return@launch
                 }
                 if (user != null) {
-                    validateIndexPermissions(0, indicesToAdd.values.toList())
-                } else {
-                    removeClosedIndices()
+                    withClosableContext(IndexManagementSecurityContext("AddPolicyHandler", Settings.EMPTY, threadContext!!, user)) {
```
bowenlan-amzn, line 151: Just input `client.threadPool().threadContext` here

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/transport/action/addpolicy/TransportAddPolicyAction.kt
```diff
@@ -145,55 +148,36 @@ class TransportAddPolicyAction @Inject constructor(
...
-                            }
-                        }
-                    }
+        private suspend fun validateIndexPermissions(indices: List<String>) {
+            indices.forEach { index ->
+                try {
+                    val ack: AcknowledgedResponse =
+                        client.suspendUntil { execute(ManagedIndexAction.INSTANCE, ManagedIndexRequest().indices(index), it) }
+                    permittedIndices.add(index)
+                    log.debug("AddPolicy ack response: [${ack.isAcknowledged}]")
```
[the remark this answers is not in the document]
petardz: Actually this will either throw or return ack=true. I think I should remove saving and logging response?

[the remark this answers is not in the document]
bowenlan-amzn: Yes, can remove

src/main/kotlin/org/opensearch/indexmanagement/opensearchapi/OpenSearchExtensions.kt
```diff
@@ -290,7 +290,12 @@ class IndexManagementSecurityContext(
      */
     override fun updateThreadContext(context: CoroutineContext) {
         logger.debug("Setting security context in thread ${Thread.currentThread().name} for job $id")
-        injector.injectRoles(if (user == null) DEFAULT_INJECT_ROLES else user.roles)
+        if (user == null) {
+            injector.injectRoles(DEFAULT_INJECT_ROLES)
+        } else {
+            injector.injectRoles(user.roles)
+            injector.injectUser(user.name)
```
petardz, line 297: I had inject user in context for index permissions(patterns) to be verified by Security plugin. This information being tied to a role, I would expect by just injecting roles, that it would be enough.

 @bowenlan-amzn

cwperks: It should not be necessary to call on inject user directly. The security plugin also does not support injecting user directly into the thread context by default, this is a feature that is turned off and the setting to enable the feature makes it clear that it is an unsupported feature: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/UserInjector.java#L62

I'm puzzled to why its not working with only using `injectRoles` - I will continue looking at this further, but here's the analysis I have done so far:

- `inject.injectRoles` injects a threadcontext headers: 1)  `opendistro_security_injected_roles` - example looks like `plugin|role1,role2`. `plugin` in this user string is a special username where the permissions check is evaluated on the `plugin` user that has the roles the plugin would like to evaluate privileges on.
- Within the security plugin it hits here - https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/filter/SecurityFilter.java#L169
- Digging into `rolesInjector.injectUserAndRoles` it populates the `opendistro_security_user` at the very end inside the `addUser` method: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/auth/RolesInjector.java#L48-L86
- Since injected roles are non-null it should bypass this block that allows local direct requests from plugins to directly pass through the SecurityFilter: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/filter/SecurityFilter.java#L258-L266


By calling on `injectUser` and passing the username it is populating a threadcontext header as `injected_user`: `<username>`

If the `plugins.security.unsupported.inject_user.enabled` is turned on inside `opensearch.yml` then it would hit this block in the [SecurityFilter](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/filter/SecurityFilter.java#L171-L179), but only if the `opendistro_security_user` transient header is null (which it shouldn't be after a roles injection) and the feature is enabled.

@petardz What list of roles are you testing with?

petardz: This is where we create role and assign role to user: [link](https://github.com/opensearch-project/index-management/pull/779/files#diff-fdccc9aa2769e6b2207f3965f104de98d2e9dd20d6059554e202c6f7ef417b98R61) and implementations of those util functions: [link](https://github.com/petardz/index-management/blob/recursion-removal/src/test/kotlin/org/opensearch/indexmanagement/SecurityRestTestCase.kt#L420)

cwperks: Could it have been hitting this block in [InjectSecurity](https://github.com/opensearch-project/common-utils/blob/main/src/main/java/org/opensearch/commons/InjectSecurity.java#L163-L165) where it does nothing if the list of roles is empty?

petardz: As of today I cannot reproduce this issue anymore... Two days ago I was flipping between "with injectUser" and "without injectUser" and was able to reproduce it every time.. My testing setup was official 2.7 docker image with installed security plugin where I installed ISM through script.

Thanks for looking into this and sorry for false alarm. I'll revert my change regarding threadContext

cwperks: @petardz Were you testing locally or via a Github action? If there's a github action log available, could you provide a link? I'm glad its working now :)

petardz: @cwperks It was locally tested with my own scripts.

src/test/kotlin/org/opensearch/indexmanagement/PolicySecurityBehaviorIT.kt
```diff
@@ -0,0 +1,124 @@
...
+            createPolicy(policy, policy.id, true, client())
+            // Call AddPolicyAction as user
+            addPolicyToIndex(index = allIndicesJoined, policyId = policy.id, expectedStatus = RestStatus.OK, client = ismUserClient!!)
+
+            refreshAllIndices()
+
+            val searchResponse = responseAsMap(client().makeRequest(RestRequest.Method.GET.toString(), "$INDEX_MANAGEMENT_INDEX/_search?size=1000"))
+            val numOfHits = ((searchResponse["hits"] as Map<*, *>)["total"] as Map<*, *>)["value"] as Int
+            // 1 Policy document + 5 ManagedIndex docs
+            assertEquals(1 + 5, numOfHits)
```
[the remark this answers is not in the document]
petardz: done

bowenlan-amzn (approved):

getsaurabh02 (approved):

bowenlan-amzn:
@petardz pls do a manual backport... not sure what's conflict that keeps failing the auto backport

PR #789, opened 2023-05-24
Support for system index interface

bowenlan-amzn:
*Issue #, if available:*

*Description of changes:*

*CheckList:*


src/main/kotlin/org/opensearch/indexmanagement/opensearchapi/OpenSearchExtensions.kt
```diff
@@ -181,6 +182,12 @@ suspend fun <T> BackoffPolicy.retry(
             } else {
                 throw e
             }
+        } catch (rje: OpenSearchRejectedExecutionException) {
+            if (iter.hasNext()) {
```
Hailong-am, line 186: should we add a check on flag `rje.isExecutorShutdown` to make sure executor is not going to shutdown?

bowenlan-amzn: Didn't notice this flag, after a look, I think it's ok to just retry on this exception since I cannot think what different thing we could do when the executor is shut down.

Hailong-am: If executer going to shutdown, retry may not work and it will get rejected again. But anyway, after i searched all `OpenSearchRejectedExecutionException` in core, there is no case this flag is set to `true` we can skip the check for now.

src/test/kotlin/org/opensearch/indexmanagement/rollup/resthandler/RestStartRollupActionIT.kt
```diff
@@ -247,6 +247,8 @@ class RestStartRollupActionIT : RollupRestTestCase() {
         assertEquals("Rollup is not RETRY", RollupMetadata.Status.RETRY, rollupMetadata.status)
 
         // clearing the config index to prevent other tests using this multi shard index
+        Thread.sleep(2000L)
```
getsaurabh02, line 250: Why do we need to wait before deletion. Also adding sleeps in general tend to make tests flaky.

src/main/kotlin/org/opensearch/indexmanagement/opensearchapi/OpenSearchExtensions.kt
```diff
@@ -181,6 +182,12 @@ suspend fun <T> BackoffPolicy.retry(
...
+                backoff = iter.next()
+                logger.warn("Rejected execution. Retrying in $backoff.", rje)
+                delay((backoff.millis))
```
getsaurabh02, line 189: Shall we add test to verify the behavior for this exception handling?

src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
```diff
@@ -77,6 +77,11 @@ abstract class IndexManagementRestTestCase : ODFERestTestCase() {
 
     protected fun Response.restStatus(): RestStatus = RestStatus.fromCode(this.statusLine.statusCode)
 
+    protected fun isIndexExists(index: String): Boolean {
```
getsaurabh02, line 80: Can we also assert on System Index property here to ensure it always remains a system index?

src/test/kotlin/org/opensearch/indexmanagement/rollup/resthandler/RestStartRollupActionIT.kt
bowenlan-amzn: Before it's to wait for rollup metadata update to finish, observe the metadata update may recreate the index after deletion. But cannot guarantee it's the root cause.
This flaky is not related to the change in this PR.

src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
bowenlan-amzn: This is a REST API call.
It seems there's no way to check whether an index is system index using REST API.
https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/indices/SystemIndices.java

src/main/kotlin/org/opensearch/indexmanagement/opensearchapi/OpenSearchExtensions.kt
bowenlan-amzn: Don't feel there's need to check such simple logic 😛 no test for the existing block also

getsaurabh02: Its good to have unit tests for code coverage and setting the code expectation for future (or accidental) changes.

getsaurabh02 (approved):

praveensameneni (approved):

PR #870, opened 2023-07-22
fix for max & min aggregations when no metric property exist

sbcd90:
*Issue #, if available:*

*Description of changes:*
fix for max & min aggregations when no metric property exist

*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/rollup/RollupIndexer.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/rollup/RollupIndexer.kt
@@ -116,8 +116,9 @@ class RollupIndexer(
             it.aggregations.forEach {
                 when (it) {
                     is InternalSum -> aggResults[it.name] = it.value
-                    is InternalMax -> aggResults[it.name] = it.value
-                    is InternalMin -> aggResults[it.name] = it.value
+                    // TODO: Need to redo the logic in corresponding doXContentBody of InternalMax and InternalMin
+                    is InternalMax -> if (it.value.isInfinite()) aggResults[it.name] = null else aggResults[it.name] = it.value
+                    is InternalMin -> if (it.value.isInfinite()) aggResults[it.name] = null else aggResults[it.name] = it.value
                     is InternalValueCount -> aggResults[it.name] = it.value
                     is InternalAvg -> aggResults[it.name] = it.value
                     else -> error("Found aggregation in composite result that is not supported [${it.type} - ${it.name}]")
--- a/src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
+++ b/src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
@@ -153,6 +153,11 @@ abstract class IndexManagementRestTestCase : ODFERestTestCase() {
         insertSampleBulkData(index, javaClass.classLoader.getResource("data/nyc_5000.ndjson").readText())
     }
 
+    protected fun generateMessageLogsData(index: String = "message-logs") {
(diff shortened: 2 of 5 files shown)


src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
```diff
@@ -152,6 +152,11 @@ abstract class IndexManagementRestTestCase : ODFERestTestCase() {
...
...ore_above":10000,"type":"keyword"},"tty_terminal":{"ignore_above":10000,"type":"keyword"},"used_hint":{"type":"boolean"},"user_group_id":{"ignore_above":10000,"type":"keyword"},"user_policy":{"ignore_above":10000,"type":"keyword"},"user_realm":{"ignore_above":10000,"type":"keyword"},"user_resource":{"ignore_above":10000,"type":"keyword"},"user_resource_type":{"ignore_above":10000,"type":"keyword"},"user_role":{"ignore_above":10000,"type":"keyword"},"user_type":{"ignore_above":10000,"type":"keyword"},"uuid":{"ignore_above":10000,"type":"keyword"},"virtual_source_address":{"ignore_above":10000,"type":"keyword"},"virtual_source_name":{"ignore_above":10000,"type":"keyword"},"was_fuzzied":{"type":"boolean"},"was_guessed":{"type":"boolean"},"watchlist":{"ignore_above":10000,"type":"keyword"},"wireless_ap":{"ignore_above":10000,"type":"keyword"},"wireless_bssid":{"ignore_above":10000,"type":"keyword"},"wireless_channel":{"ignore_above":10000,"type":"keyword"},"wireless_encryption":{"ignore_above":10000,"type":"keyword"},"wireless_ssid":{"ignore_above":10000,"type":"keyword"},"x_att_tenantid":{"ignore_above":10000,"type":"keyword"}}},"timeStamp":{"ignore_above":10000,"type":"keyword"}}""")
```
bowenlan-amzn, line 156: Wondering why putting so many fields here? same for the sample data

sbcd90: hi @bowenlan-amzn , updated the mappings to only have necessary fields & updated sample data too.

src/main/kotlin/org/opensearch/indexmanagement/rollup/RollupIndexer.kt
```diff
@@ -116,8 +116,8 @@ class RollupIndexer(
             it.aggregations.forEach {
                 when (it) {
                     is InternalSum -> aggResults[it.name] = it.value
-                    is InternalMax -> aggResults[it.name] = it.value
-                    is InternalMin -> aggResults[it.name] = it.value
+                    is InternalMax -> if (it.value.isInfinite()) aggResults[it.name] = null else aggResults[it.name] = it.value
```
bowenlan-amzn, line 119: The reason this was missed before is because the value of max/min in InternalMax/InternalMin is initialized in method `reduce`, and processed in `doXContentBody` before responding. However, the it.value we got here has not been processed by `doXContentBody`. This is easy to miss, can we add a comment here in case in the future we are adding more metrics and may miss again. like `Need to redo the logic in corresponding doXContentBody` https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/aggregations/metrics/InternalMax.java#L101

sbcd90: hi @bowenlan-amzn , i added a comment.

bowenlan-amzn (approved):

AWSHurneyt (approved):

Issue #734, opened 2023-04-09
[BUG] index rollover - doesn't create all aliases in the newly created index


oridool:
**What is the bug?**
When using ISM rollover and the index contains multiple aliases, only the alias configured for rollover is created in the new index.
Other aliases existing in original index (before rollover) are not copied.

**How can one reproduce the bug?**

1. create ism policy and index template such as:
```
PUT  _index_template/rollovertests_template
{
  "index_patterns": [
    "rollovertests-*"
  ],
  "template": {
    "settings": {
      "number_of_shards": 1,
      "number_of_replicas": 1,
      "plugins.index_state_management.rollover_alias": "rolloveralias"
    },
    "mappings": {
      "_routing": {
      "required": true
      },
      "properties": {
        "timestamp": {
          "type": "date",
          "format": "yyyy-MM-dd HH:mm:ss||yyyy-MM-dd||epoch_millis"
        }
      }
    }
  }
}

PUT _plugins/_ism/policies/rollovertests_policy
{
  "policy": {
    "description": "test rollover policy.",
    "default_state": "rollover",
    "states": [
      {
        "name": "rollover",
        "actions": [
          {
            "rollover": {
              "min_index_age": "2m"
            }
          }
        ],
        "transitions": []
      }
    ] 
    ,
    "ism_template": {
      "index_patterns": ["rollovertests*"],
      "priority": 100
    }
  }
}

```
2. create index with more than one alias:
```
PUT rollovertests-000001
{
  "aliases": {
    "rolloveralias": {
      "is_write_index": true,
      "filter": {
        "term": {
          "timestamp": "1574641891142"
        }
      },
      "routing": "12345"
    },
    "anotheralias": {
      "filter": {
        "term": {
          "timestamp": "157464189133"
        }
      },
      "routing": "12345"
    }
  }
}
```
3.  After rollover, check the newly created index:
```
GET rollovertests-000002
```
Result:
```
{
  "rollovertests-000002": {
    "aliases": {
      "rolloveralias": {
        "filter": {
          "term": {
            "timestamp": "1574641891142"
          }
        },
        "index_routing": "12345",
        "search_routing": "12345",
        "is_write_index": true
      }
    },...

``` 

As you can see, the alias 'anotheralias' is missing from the new index 'rollovertests-000002'.

**What is the expected behavior?**

The alias 'anotheralias' should be copied to the new index upon rollover. 
In fact, all existing aliases with all of their properties (routing, filter,...) should be copied.
[...]

Issue #849, opened 2023-07-10
[FEATURE] Copy metadata from last index during rollover


soosinha:
**Is your feature request related to a problem?**
Follow up from the feature request in OpenSearch: https://github.com/opensearch-project/OpenSearch/issues/8419

During rollover, the index metadata is picked up from the template. But in case the mappings have been updated during ingestion due to dynamic mappings being enabled, the new index will not have the latest mappings. When data is ingested on the new index, there will be updation of mappings immediately which will result in cluster state updates. These cluster state updates can potentially increase the load on the cluster manager nodes. 

**What solution would you like?**
Rollover action should have an option to copy the metadata from the previous index. This way the mappings will be in sync with documents being ingested and we can avoid dynamic mappings updates.

Detailed API level changes:
The rollover action will have a new boolean parameter `use_previous_index_metadata` like below:
```
{
  "rollover": {
    "use_previous_index_metadata": true
  }
}
```
This parameter will be false by default. When this is set to true, during rollover the metadata will be copied from the last write index of the alias or the data stream.

soosinha:
@opensearch-project/index-management Let me know if I can go ahead with this change.

bowenlan-amzn:
@soosinha please go ahead. 
We have [a request](https://github.com/opensearch-project/index-management/issues/734) to copy over all aliases from previous index to newly created index during rollover, will your enhancement also cover aliases or it will only cover mappings of index metadata.

bowenlan-amzn:
Thanks a lot! 
I would suggest using separate parameters for aliases, settings and mappings.

JahanviAcquia:
any update on this ticket? We have dynamic attribute mapping. So we will need to update rollover alias manually

PR #892, opened 2023-08-15
Support copy alias in rollover

bowenlan-amzn:
*Issue #, if available:*
#734 

*Description of changes:*

- Introduce a new parameter `copy_alias` under rollover action, that can let user choose whether to copy the alias from old index to new index during rollover. 
  - This param is default to false because rollover API by default doesn't copy over alias.
- If enabled, copy alias will only be performed after rollover succeeded. 
- If copy alias failed, it can be retried base on the `rolled_over_index_name` saved in metadata. `rolled_over_index_name` is saved after rollover succeeded
- New fields `copy_alias`, `rolled_over_index_name` are added to the ISM config index mapping, making sure it won't be added by dynamic mapping and causes mapping conflict regression
  - `copy_alias` in policy.action will be handled during create/update policy path. The policy field in managed_index is of object type so no problem there.
  - `rolled_over_index_name` in metadata will be handled during updateMetadata in Runner, the mapping will first be updated then the metadata document will be saved.
- Add BWC to make sure it won't cause rollover job from old versions to fail. 
1. Start cluster using the old ISM, create a rollover job
2. Upgrade the cluster using the new ISM, trigger the rollover, double check the managed_index and metadata content

*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
@@ -7,12 +7,18 @@ package org.opensearch.indexmanagement.indexstatemanagement.step.rollover
 
 import org.apache.logging.log4j.LogManager
 import org.opensearch.ExceptionsHelper
+import org.opensearch.action.admin.indices.alias.IndicesAliasesRequest
+import org.opensearch.action.admin.indices.alias.IndicesAliasesRequest.AliasActions
 import org.opensearch.action.admin.indices.rollover.RolloverRequest
 import org.opensearch.action.admin.indices.rollover.RolloverResponse
 import org.opensearch.action.admin.indices.stats.IndicesStatsRequest
 import org.opensearch.action.admin.indices.stats.IndicesStatsResponse
 import org.opensearch.common.unit.ByteSizeValue
+import org.opensearch.action.support.master.AcknowledgedResponse
+import org.opensearch.client.Client
+import org.opensearch.cluster.service.ClusterService
 import org.opensearch.common.unit.TimeValue
+import org.opensearch.index.IndexNotFoundException
 import org.opensearch.indexmanagement.indexstatemanagement.action.RolloverAction
 import org.opensearch.indexmanagement.indexstatemanagement.opensearchapi.getRolloverAlias
 import org.opensearch.indexmanagement.indexstatemanagement.opensearchapi.getRolloverSkip
@@ -33,6 +39,7 @@ class AttemptRolloverStep(private val action: RolloverAction) : Step(name) {
     private val logger = LogManager.getLogger(javaClass)
     private var stepStatus = StepStatus.STARTING
     private var info: Map<String, Any>? = null
+    private var newIndex: String? = null // this variable holds the new index name if rollover is successful in this run
 
     @Suppress("ComplexMethod", "LongMethod")
     override suspend fun execute(): Step {
@@ -53,6 +60,9 @@ class AttemptRolloverStep(private val action: RolloverAction) : Step(name) {
         if (clusterService.state().metadata.index(indexName).rolloverInfos.containsKey(rolloverTarget)) {
             stepStatus = StepStatus.COMPLETED
             info = mapOf("message" to getAlreadyRolledOverMessage(indexName, rolloverTarget))
+
+            // If already rolled over, alias may not get copied over yet
+            copyAlias(clusterService, indexName, context.client, rolloverTarget, context.metadata)
             return this
         }
 
@@ -114,6 +124,7 @@ class AttemptRolloverStep(private val action: RolloverAction) : Step(name) {
                     " numDocs=$numDocs, indexSize=${indexSize.bytes}, primaryShardSize=${largestPrimaryShardSize.bytes}]"
             )
             executeRollover(context, rolloverTarget, isDataStream, conditions)
+            copyAlias(clusterService, indexName, context.client, rolloverTarget, context.metadata)
         } else {
             stepStatus = StepStatus.CONDITION_NOT_MET
             info = mapOf("message" to getPendingMessage(indexName), "conditions" to conditions)
@@ -158,14 +169,14 @@ class AttemptRolloverStep(private val action: RolloverAction) : Step(name) {
         val indexName = context.metadata.index
         val metadata = context.clusterService.state().metadata
         val indexAlias = metadata.index(indexName)?.aliases?.get(alias)
-        logger.debug("Index $indexName has aliases $indexAlias")
+        logger.debug("Index {} has aliases {}", indexName, indexAlias)
         if (indexAlias == null) {
             return false
         }
         val isWriteIndex = indexAlias.writeIndex() // this could be null
         if (isWriteIndex != true) {
             val aliasIndices = metadata.indicesLookup[alias]?.indices?.map { it.index }
-            logger.debug("Alias $alias contains indices $aliasIndices")
+            logger.debug("Alias {} contains indices {}", alias, aliasIndices)
             if (aliasIndices != null && aliasIndices.size > 1) {
                 return false
             }
@@ -225,6 +236,9 @@ class AttemptRolloverStep(private val action: RolloverAction) : Step(name) {
                     else -> getSuccessMessage(indexName)
                 }
 
+                // Save newIndex later to metadata to be reused in case of failures
+                newIndex = response.newIndex
+
                 stepStatus = StepStatus.COMPLETED
                 info = listOfNotNull(
                     "message" to message,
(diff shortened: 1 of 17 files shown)


eirsep:
why is copyAlias defaulting to false. it seems like a very useful feature.

eirsep:
have we done any upgrade testing to ensure we are not introducing regression in mapping to verify ISM config index mapping change works as expected?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -288,5 +385,13 @@ class AttemptRolloverStep(private val action: RolloverAction) : Step(name) {
         fun getSkipRolloverMessage(index: String) = "Skipped rollover action for [index=$index]"
         fun getAlreadyRolledOverMessage(index: String, alias: String) =
             "This index has already been rolled over using this alias, treating as a success [index=$index, alias=$alias]"
+        fun getSuccessCopyAliasMessage(index: String, newIndex: String) =
+            "Successfully rolled over and copied alias from [index=$index] to [index=$newIndex]"
+        fun getFailedCopyAliasMessage(index: String, newIndex: String) =
+            "Successfully rolled over but failed to copied alias from [index=$index] to [index=$newIndex]"
+        fun getCopyAliasIndexNotFoundMessage(newIndex: String?) =
+            "Successfully rolled over but new index [index=$newIndex] not found during copy alias"
+        fun getCopyAliasRolledOverIndexNotFoundMessage(index: String?) =
```
eirsep, line 394: can we add test for this scenario .. coverage has dropped with PR

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -251,21 +265,104 @@ class AttemptRolloverStep(private val action: RolloverAction) : Step(name) {
...
             stepMetaData = StepMetaData(name, getStepStartTime(currentMetadata).toEpochMilli(), stepStatus),
             rolledOver = if (currentMetadata.rolledOver == true) true else stepStatus == StepStatus.COMPLETED,
+            rolledOverIndexName = if (currentMetadata.rolledOverIndexName != null) currentMetadata.rolledOverIndexName else newIndex,
             transitionTo = null,
             info = info
         )
     }
 
-    private fun handleException(indexName: String, e: Exception, message: String = getFailedMessage(indexName)) {
+    private fun handleException(indexName: String, e: Exception, message: String = getFailedMessage(indexName), conditions: Any? = null) {
```
eirsep, line 359: can we add test for this scenario .. coverage has dropped with PR

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/RolloverActionIT.kt
```diff
@@ -632,4 +630,102 @@ class RolloverActionIT : IndexStateManagementRestTestCase() {
         }
         assertTrue("New rollover index does not exist.", indexExists("$indexNameBase-000002"))
     }
+
+    @Suppress("UNCHECKED_CAST")
+    fun `test rollover with copy alias`() {
```
eirsep, line 635: we should also add a test case where copy alias fails.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -251,21 +265,104 @@ class AttemptRolloverStep(private val action: RolloverAction) : Step(name) {
...
+            stepStatus = StepStatus.COMPLETED
+            info = listOfNotNull(
+                "message" to getCopyAliasRolledOverIndexNotFoundMessage(indexName),
+                if (conditions != null) "conditions" to conditions else null
+            ).toMap()
+            return
+        }
+
+        val aliasActions = mutableListOf<AliasActions>()
+        val aliases = clusterService.state().metadata().index(indexName).aliases
```
sbcd90, line 304: can this line also throw `IndexNotFoundException`?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -251,21 +265,104 @@ class AttemptRolloverStep(private val action: RolloverAction) : Step(name) {
...
+        if (!action.copyAlias) return
+
+        // Try to preserve the rollover conditions
+        val conditions = info?.get("conditions") ?: context?.metadata?.info?.get("conditions")
+
+        val rolledOverIndexName = newIndex ?: metadata.rolledOverIndexName
+        if (rolledOverIndexName == null) {
+            // Only in rare case when the program shut down unexpectedly before rolledOverIndexName is set or metadata corrupted
+            // ISM cannot auto recover from this case, so the status is COMPLETED
+            logger.error("Cannot find rolled over index to copy aliases to")
```
eirsep, line 294: plz log some info from metadata to understand source index or any policy related info 

reading this log in log file won't shed light on which indices's metadata we are dealing with

bowenlan-amzn: Added the index here
However, we do have log that helps identify step execution already

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
bowenlan-amzn: Should not, this Step is called during ISM running on certain index, if index is deleted, ISM won't run for that then.

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/RolloverActionIT.kt
bowenlan-amzn: For the failed cases where alias API throw exception, we don't have an way to mock that in IT in ISM code base. Should be sth we can look into later.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
bowenlan-amzn: Added some UTs

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -251,21 +265,104 @@ class AttemptRolloverStep(private val action: RolloverAction) : Step(name) {
...
+        try {
+            val aliasRes: AcknowledgedResponse = client.admin().indices().suspendUntil { aliases(aliasReq, it) }
+            if (aliasRes.isAcknowledged) {
+                stepStatus = StepStatus.COMPLETED
+                info = listOfNotNull(
+                    "message" to getSuccessCopyAliasMessage(indexName, rolledOverIndexName),
+                    if (conditions != null) "conditions" to conditions else null
+                ).toMap()
+            } else {
+                stepStatus = StepStatus.FAILED
```
eirsep, line 331: if we mark this failed after rollover is completed it would send a wrong message right.
what failed here is the copyalias step which should be logged and somehow notified to user. 
Or can we have a different message that 
"rollover completed but copy alias failed"

How will user handle this by himself?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -251,21 +265,104 @@ class AttemptRolloverStep(private val action: RolloverAction) : Step(name) {
...
+                info = listOfNotNull(
+                    "message" to getFailedCopyAliasMessage(indexName, rolledOverIndexName),
```
eirsep, line 333: failure messages should also have exception cause message else we would not know the cause for failure just by looking at the policy failure

src/main/resources/mappings/opendistro-ism-history.json
```diff
@@ -1,6 +1,6 @@
 {
   "_meta" : {
-    "schema_version": 5
+    "schema_version": 6
   },
```
eirsep, line 4: is copy_alias value not required here?
would it be required to audit when it was changed from false to true?

eirsep:
Shouldn't we make this change in core and add option to copy aliases in rollover api?

src/main/resources/mappings/opendistro-ism-history.json
bowenlan-amzn: History only contains metadata

We don't support audit on policy change

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
```diff
@@ -288,5 +385,13 @@ class AttemptRolloverStep(private val action: RolloverAction) : Step(name) {
         fun getSkipRolloverMessage(index: String) = "Skipped rollover action for [index=$index]"
         fun getAlreadyRolledOverMessage(index: String, alias: String) =
             "This index has already been rolled over using this alias, treating as a success [index=$index, alias=$alias]"
+        fun getSuccessCopyAliasMessage(index: String, newIndex: String) =
+            "Successfully rolled over and copied alias from [index=$index] to [index=$newIndex]"
+        fun getFailedCopyAliasMessage(index: String, newIndex: String) =
```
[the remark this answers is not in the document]
bowenlan-amzn: Exception message is in the cause field of info

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
bowenlan-amzn: Cx can copy by themselves, uses retry API to move to next state, or let this rollover step auto retry.

We don't support running 2 steps continuously right now, so copy alias is not implemented as a separate step. If separate, it will only be done in the next job run but the requirement is to do it right after rollover.

bowenlan-amzn:
> Shouldn't we make this change in core and add option to copy aliases in rollover api?

Of course it's best to support this in core and the open issue is here https://github.com/opensearch-project/index-management/issues/849
Now we are delivering this feature ASAP for the ISM customer

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/rollover/AttemptRolloverStep.kt
bowenlan-amzn: Modify the message to indicate the copy alias call is not ack

eirsep (approved):
thanks for the changes, @bowenlan-amzn

eirsep:
Hope users can update existing ISM policies and set copyAlias=true

sbcd90 (approved):

PR #901, opened 2023-08-22
Rollup Search API : Searching both historical rollup and non-rollup data

ronnaksaxena:
This feature supports searching both historical rollup and non-rollup data
To use the endpoint is `GET <rollup_index>,<live_index1>, <live_index2>, <live_indexX>/_search`

Currently supports:

1. Sum aggregations
2. Min aggregations
3. Max aggregations
4. Avg aggregations
5. Value Count aggregations

It works like this:
overlapCodeFlow

The type of use cases that could arise are:
Base Case:  No data overlap between rollup indices and live data indices

GET /rollup-data,live-data/_search
noOverlap

Case 2: Overlapping Data Between 1 Rollup index and 1 Live index
Attempting to handle this use case from my design doc:
1Overlap

Flow for this code would look like:

case2FlowChart

Case 3: Overlapping data between multiple (aliased) live data indices 
Can add a comma separated list of all index names or index alias/template
`GET /rollup-data,live-data-alias/_search`
multiLIveData
Screenshot 2023-08-30 at 2 55 42 PM

*CheckList:*
- [x ] Commits are signed per the DCO using --signoff


src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
```diff
@@ -6,56 +6,69 @@
...
 import org.opensearch.transport.TransportRequestHandler
-
+import java.nio.ByteBuffer
+import java.util.Base64
 class RollupInterceptor(
     val clusterService: ClusterService,
     val settings: Settings,
-    val indexNameExpressionResolver: IndexNameExpressionResolver
+    val indexNameExpressionResolver: IndexNameExpressionResolver,
+    val xContentRegistry: NamedXContentRegistry
```
bowenlan-amzn, line 71: Don't see this being used in this file?

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
```diff
@@ -71,6 +84,93 @@ class RollupInterceptor(
...
+            for (agg in currentAggs) {
+                if (agg.name == "interceptor_interval_data") {
+                    return true
+                }
+            }
+        }
+        return false
+    }
+    // Helper fn to avoid rewritting a rollup request an extra time
+    fun isReqeustRollupFormat(request: ShardSearchRequest): Boolean {
```
bowenlan-amzn, line 113: isRequest

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
```diff
@@ -71,6 +84,93 @@ class RollupInterceptor(
...
+        for (dim in rollupJob.dimensions) {
+            if (dim is DateHistogram) {
+                dateSourceField = dim.sourceField
+                rollupInterval = dim.fixedInterval!!
+                break
+            }
+        }
+        // Wraps all existing aggs in bucket aggregation
+        // Notifies the response interceptor that was rewritten since agg name is interceptor_interval_data
+        // Edge case if User selected this as the aggregation name :/
```
bowenlan-amzn, line 164: You can add random string to avoid this edge case

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
```diff
@@ -82,38 +182,65 @@ class RollupInterceptor(
...
-                    if (isRollupIndex) {
-                        if (request.source().size() != 0) {
-                            throw IllegalArgumentException("Rollup search must have size explicitly set to 0, but found ${request.source().size()}")
+                    val (containsRollup, rollupJob) = originalSearchContainsRollup(request)
+                    val shardRequestIndex = request.shardId().indexName
+                    val isRollupIndex = isRollupIndex(shardRequestIndex, clusterService.state())
+                    // Only modifies rollup searches and avoids internal client calls
+                    if (containsRollup || isRollupIndex) {
+                        val (concreteRollupIndicesArray, concreteLiveIndicesArray) = getConcreteIndices(request)
+                        // Avoid infinite interceptor loop
```
bowenlan-amzn, line 191: Since no client call in request interceptor now, we won't need to avoid infinite loop right

On the other hand, since you have a way to avoid infinite loop now, what would be the pros and cons to do client call in request interceptor side compared to response interceptor side?

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
```diff
@@ -0,0 +1,423 @@
...
+                    }
+                }
+                // when just 1 rollup index is in request, keep for testing
+                is QueryFetchSearchResult -> {
+                    val queryResult = response.queryResult()
+                    if (queryResult.hasAggs() && isRewrittenInterceptorRequest(queryResult)) {
+                        // Check for overlap
+                        val (startTime, endTime) = findOverlap(queryResult)
+                        // Modify agg to be original result without overlap computed in
+                        queryResult.aggregations(computeAggregationsWithoutOverlap(queryResult.aggregations().expand(), startTime, endTime))
```
bowenlan-amzn, line 398: Seems you are dealing with overlap in both live and rollup response, wondering should you only remove the result from one?

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
```diff
@@ -205,63 +205,70 @@ class ResponseInterceptor(
...
                     }
 
                     override fun onFailure(e: Exception) {
-                        logger.error("ronsax minLiveDate request failed ", e)
+                        logger.error("ronsax maxLiveDate request failed ", e)
                         latch.countDown()
                     }
                 }
             )
             latch.await()
```
eirsep, line 239: plz replace countdown latch usage with kotlin coroutines.

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
```diff
@@ -0,0 +1,530 @@
...
+                }
+            }
+            return InternalAggregations(allAggregations, null)
+        }
+        @Suppress("UNCHECKED_CAST")
+        override fun handleResponse(response: T?) {
+            // Handle the response if it came from interceptor
+            when (response) {
+                // live index
+                is QuerySearchResult -> {
```
bowenlan-amzn, line 487: Not sure why live index is QuerySearchResult, but for rollup index is QueryFetchSearchResult, why the difference here?

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
```diff
@@ -0,0 +1,530 @@
...
+            val (rollupIndices, liveIndices) = getRollupAndLiveIndices(request)
+            val shardRequestIndex = request.shardId().indexName
+            val isShardIndexRollup = isRollupIndex(shardRequestIndex, clusterService.state())
+            // Build search request to find the maximum date on the rolled data index
+            var sort = SortBuilders.fieldSort("$dateTargetField.date_histogram").order(SortOrder.DESC)
+            var searchSourceBuilder = SearchSourceBuilder()
+                .sort(sort)
+                .query(oldQuery)
+                .size(1)
+            // Need to avoid infinite interceptor loop
```
bowenlan-amzn, line 257: Comment here how do you avoid the infinite loop

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
```diff
@@ -0,0 +1,530 @@
...
+            // Build search request to find the minimum date in the live data index
+            sort = SortBuilders.fieldSort(dateSourceField).order(SortOrder.ASC)
+            searchSourceBuilder = SearchSourceBuilder()
+                .sort(sort)
+                .size(1)
+            val minLiveDateRequest = SearchRequest()
+                .source(searchSourceBuilder)
+            /*
+            If the shard index is a rollup index I want to find the minimum value of all the live indices to compute the overlap
+            If the shard index is a live index I only care about the minimum value of the current shard index
```
bowenlan-amzn, line 287: Some explanation would help

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
```diff
@@ -0,0 +1,530 @@
...
+                        maxRolledDateResponse = searchResponse
+                        latch.countDown()
+                    }
+
+                    override fun onFailure(e: Exception) {
+                        logger.error("ronsax maxLiveDate request failed ", e)
+                        latch.countDown()
+                    }
+                }
+            )
```
bowenlan-amzn, line 276: We don't want this client call to be executed for each shard response

So investigate if you can save the result of this call to a variable in this class ResponseInterceptor, and reuse for other shards response

getsaurabh02 (commented):
Lets add unit tests at the class level to be able to verify helper/util and class methods

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
ronnaksaxena: It doesn't work with coroutines

src/main/kotlin/org/opensearch/indexmanagement/rollup/util/RollupUtils.kt
```diff
@@ -62,6 +62,7 @@ import org.opensearch.search.aggregations.metrics.ScriptedMetricAggregationBuild
 import org.opensearch.search.aggregations.metrics.SumAggregationBuilder
 import org.opensearch.search.aggregations.metrics.ValueCountAggregationBuilder
 import org.opensearch.search.builder.SearchSourceBuilder
+import java.util.*
```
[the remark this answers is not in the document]
ronnaksaxena: removed!

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
ronnaksaxena: Sorry that was for a helper function I removed. Will delete it!

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
ronnaksaxena: This function check if the request was rewritten to match the rollup index data. I do not think isRequest is descriptive enough for its purpose.

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
ronnaksaxena: It is more of a design decision to add the client to the response interceptor. There are no advantages to either.

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
ronnaksaxena: Design decision to always include all data from live indices. Consider this graphic on the differences between the two: 
midIntervalOverlap
Untitled

ronnaksaxena: maybe a future optimization to allow the user to decide which option they would prefer?

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
```diff
@@ -82,38 +220,65 @@ class RollupInterceptor(
...
+                        if (concreteRollupIndicesArray.isNotEmpty() && concreteLiveIndicesArray.isNotEmpty() && request.source().aggregations() != null && !isRequestRewrittenIntoBuckets(request)) {
+                            // Break apart request to remove overlapping parts
+                            breakRequestIntoBuckets(request, rollupJob!!)
```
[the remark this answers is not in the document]
ronnaksaxena: @bowenlan-amzn I add another aggregation but wrap the existing aggregations into a sub aggregation. Would a more appropriate name be putAggregationsIntoBuckets() ?

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
ronnaksaxena: Removed this. I kept it for testing by sending a request with the aggregation name as "interceptor_interval_data" I can check if the behavior in the response interceptor is correct. (The response would be a QueryFetchSearchReuslt) on one rollup index)

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
ronnaksaxena: I added a comment, let me know if I need more clarification

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
ronnaksaxena: Added!

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
ronnaksaxena: Will look into this if I am time. I think a core modification will be needed.

bowenlan-amzn:
Now we can see this change broken data stream scenario https://github.com/opensearch-project/index-management/actions/runs/6125162603/job/16626572043?pr=901

```
Tests with failures:
 - org.opensearch.indexmanagement.rollup.runner.RollupRunnerIT.test metadata is created for data stream rollup job when none exists
 - org.opensearch.indexmanagement.transform.TransformRunnerIT.test transform with data stream
 - ```

File: src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt [hunk 13]
```diff
@@ -195,6 +194,7 @@ class ResponseInterceptor(
                 .indices(*rollupIndices) // add all rollup indices to this request
             var maxRolledDateResponse: SearchResponse? = null
             var latch = CountDownLatch(1)
+            logger.info("ronsax sending request to find max rollup time for index: $shardRequestIndex")
```
eirsep, line 197: plz fix info log

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
```diff
@@ -230,6 +230,7 @@ class ResponseInterceptor(
 
             var minLiveDateResponse: SearchResponse? = null
             latch = CountDownLatch(1)
+            logger.info("ronsax sending request to find minLiveData for index: $shardRequestIndex")
```
eirsep, line 233: plz fix info log

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
```diff
@@ -256,6 +257,7 @@ class ResponseInterceptor(
                 // If intersection found on rollup index, remove overlap
                 if ((liveDataStartPoint < rollupDataEndPoint) && isShardIndexRollup) {
                     // Start at 0, end at live data
+                    logger.info("ronsax sending request to find rollup endtime for index: $shardRequestIndex")
```
eirsep, line 260: plz fix info log

eirsep:
Can you add TODOs on the client calls made from within the shard level interceptor as this would imply for one rollup search request, we are making n more search requests where n is number of shards and that would have performance implications.

We can merge this PR and continue to work on it.

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
```diff
@@ -0,0 +1,394 @@
...
+                logger.error("Not able to retrieve intersection time from response: $e")
+            }
+            return 0L // dummy :P
+        }
+
+//         Returns Pair(startRange: Long, endRange: Long)
+//         Note startRange is inclusive and endRange is exclusive, they are Longs because the type is epoch milliseconds
+        @Suppress("LongMethod", "SpreadOperator")
+        suspend fun findOverlap(response: QuerySearchResult): Pair<Long, Long> {
+            val job: Rollup = getRollupJob(response)!! // maybe throw a try catch later
```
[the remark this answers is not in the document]
eirsep: we don't need a todo for this. Plz add try catch right away

.github/workflows/multi-node-test-workflow.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: Checkout Branch
         uses: actions/checkout@v2
       - name: Run integration tests with multi node config
-        run: ./gradlew integTest -PnumNodes=3
+        run: ./gradlew integTest
```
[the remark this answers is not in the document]
ronnaksaxena: Sorry was an accident to debug the integ test, I added it back

AWSHurneyt (approved):

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
bowenlan-amzn: intent to say there's a typo here...

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
```diff
@@ -82,38 +190,77 @@ class RollupInterceptor(
...
-                    if (isRollupIndex) {
-                        if (request.source().size() != 0) {
-                            throw IllegalArgumentException("Rollup search must have size explicitly set to 0, but found ${request.source().size()}")
+                    val isDataStream = (request.indices().any { IndexUtils.isDataStream(it, clusterService.state()) })
+                    val shardRequestIndex = request.shardId().indexName
+                    // isRollupIndex throws an exception if the request is on a data stream
+                    val isRollupIndex = if (isDataStream) false else isRollupIndex(shardRequestIndex, clusterService.state())
+                    val (containsRollup, rollupJob) = if (isDataStream) Pair(false, null) else originalSearchContainsRollup(request)
+                    // Only modifies rollup searches and avoids internal client calls
+                    if (containsRollup || isRollupIndex) {
```
bowenlan-amzn, line 199: Do we need `|| isRollupIndex` here?

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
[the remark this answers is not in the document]
bowenlan-amzn: maybe just breakIntoBuckets

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
```diff
@@ -82,38 +190,77 @@ class RollupInterceptor(
...
+                            otherwise do not allow size > 0 for rollup indices
+                             */
+                            if (!canHaveSize(request) && request.source().size() != 0) {
+                                throw IllegalArgumentException(
+                                    "Rollup search must have size explicitly set to 0, " +
+                                        "but found ${request.source().size()}"
+                                )
+                            }
+                            rewriteRollupRequest(request, rollupJob!!, concreteRollupIndicesArray)
                         }
```
bowenlan-amzn, line 222: Can this be moved out from the parent if condition `if (containsRollup`?

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
```diff
@@ -82,38 +190,77 @@ class RollupInterceptor(
...
-                        val allMatchingRollupJobs = validateIndicies(concreteIndices, fieldMappings)
-
-                        // only rebuild if there is necessity to rebuild
-                        if (fieldMappings.isNotEmpty()) {
-                            rewriteShardSearchForRollupJobs(request, allMatchingRollupJobs)
+                        // Rewrite the request to fit rollup format if not already done previously
+                        if (isRollupIndex && !isReqeustRollupFormat(request)) {
+                            /* Client calls from the response interceptor require request bodies of 1,
+                            otherwise do not allow size > 0 for rollup indices
+                             */
```
bowenlan-amzn, line 214: Context: we need to make call to find out the min, max time of rollup and live indexes, however, this call would be intercepted here and get re-write.

So the idea is to explicitly allow this call to go through rollup interceptor w/o re-write them, so that whether it's from response interceptor or client, both should work.

maybe change canHaveSize to `allowRequest` and comment the above reason

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
```diff
@@ -0,0 +1,408 @@
...
+        return CustomAsyncSender(sender)
+    }
+
+    private inner class CustomAsyncSender(private val originalSender: TransportInterceptor.AsyncSender) : TransportInterceptor.AsyncSender {
+        override fun <T : TransportResponse?> sendRequest(
+            connection: Transport.Connection?,
+            action: String?,
+            request: TransportRequest?,
+            options: TransportRequestOptions?,
+            handler: TransportResponseHandler<T>?
```
bowenlan-amzn, line 81: Pls clean up the question marks

.github/workflows/multi-node-test-workflow.yml
```diff
@@ -3,7 +3,7 @@ name: Multi node test workflow
 on:
   pull_request:
     branches:
-      - "*"
```
eirsep, line 6: can we remove this after PR is approved.

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/ResponseInterceptor.kt
```diff
@@ -0,0 +1,408 @@
...
+import org.opensearch.transport.TransportRequestOptions
+import org.opensearch.transport.TransportResponse
+import org.opensearch.transport.TransportResponseHandler
+import java.time.ZonedDateTime
+import kotlin.math.max
+import kotlin.math.min
+
+/**
+* The Response Interceptor class modifies resopnses if the search API is triggered on rollup and live data
+* 1. The class check if the request was rewritten in the RollupInterceptor into more granular buckets
```
eirsep, line 58: NITPICK: the class checks*

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
```diff
@@ -82,38 +190,77 @@ class RollupInterceptor(
...
             override fun messageReceived(request: T, channel: TransportChannel, task: Task) {
                 if (searchEnabled && request is ShardSearchRequest) {
-                    val index = request.shardId().indexName
-                    val isRollupIndex = isRollupIndex(index, clusterService.state())
-                    if (isRollupIndex) {
-                        if (request.source().size() != 0) {
-                            throw IllegalArgumentException("Rollup search must have size explicitly set to 0, but found ${request.source().size()}")
+                    val isDataStream = (request.indices().any { IndexUtils.isDataStream(it, clusterService.state()) })
+                    val shardRequestIndex = request.shardId().indexName
+                    // isRollupIndex throws an exception if the request is on a data stream
```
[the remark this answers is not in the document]
ronnaksaxena: It is a null ptr exception when checking is the shardId indexName for a data stream is a rollup index

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
```diff
@@ -71,8 +75,112 @@ class RollupInterceptor(
             searchAllJobs = it
         }
     }
-
+    // Returns Pair<containsRollup: Boolean, rollupJob: RollupJob>
```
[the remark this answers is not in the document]
ronnaksaxena: Added!

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
ronnaksaxena: Yes this check is needed. The base case fails unless I check for both but maybe the logic could be cleaned up to have less if checks

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
ronnaksaxena: No I cannot I need the concreteLiveIndices array to call the rewriteRollupRequest helper function

src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
ronnaksaxena: Yeah that is clearer, will change!

ronnaksaxena:
@bowenlan-amzn @eirsep Thank you guys for the feedback. I pushed a new commit with suggested changes. 👍🏽

eirsep (approved):

PR #911, opened 2023-08-30
Set preference to _primary when searching control-center index

gaobinlong:
*Issue *
https://github.com/opensearch-project/index-management/issues/833

*Description of changes:*
This change is to set the query parameter `preference` to `_primary` when searching control-center index, because when indexing new documents in this index, we use RefreshPolicy.IMMEDIATE to force a refresh after the document is indexed which aims to make the new document visible immediately, but if segment replication is enabled in the control-center index, read-after-write is not guaranteed, so we need to search the primary shard only to make the document can be seen as soon as possible.

The search load on the control-center index is pretty low, it will not have some performance issue if we only search the primary shard.

*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/controlcenter/notification/action/get/TransportGetLRONConfigAction.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/controlcenter/notification/action/get/TransportGetLRONConfigAction.kt
@@ -94,7 +94,7 @@ class TransportGetLRONConfigAction @Inject constructor(
             val searchRequest = SearchRequest()
                 .source(searchSourceBuilder)
                 .indices(IndexManagementPlugin.CONTROL_CENTER_INDEX)
-                .preference(Preference.PRIMARY.type())
+                .preference(Preference.PRIMARY_FIRST.type())
 
             client.search(
                 searchRequest,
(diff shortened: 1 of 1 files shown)


src/main/kotlin/org/opensearch/indexmanagement/controlcenter/notification/action/get/TransportGetLRONConfigAction.kt
```diff
@@ -93,6 +94,7 @@ class TransportGetLRONConfigAction @Inject constructor(
             val searchRequest = SearchRequest()
                 .source(searchSourceBuilder)
                 .indices(IndexManagementPlugin.CONTROL_CENTER_INDEX)
+                .preference(Preference.PRIMARY.type())
```
dreamer-89, line 97: Thanks @gaobinlong for this change. 
`_primary` preference results in request been served via primary. With this change, we will be hitting primary shard for all indices (both docre, segrep). This would be un-necessary for `DOCUMENT` indices. Would using `_primary_first` be a deal breaker in your case ?

bowenlan-amzn:
No harm adding this. But may not be necessary IMO. 
The reasoning is the time between LRON config indexed, and needed by sending notification would be much longer than the time for segRep to sync.

bowenlan-amzn (approved):

Hailong-am (approved):

PR #983, opened 2023-10-04
Fix the flaky in allocation IT during multinode test

bowenlan-amzn:
*Issue #, if available:*

*Description of changes:*

*CheckList:*

### DIFF
--- a/src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AllocationActionIT.kt
+++ b/src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AllocationActionIT.kt
@@ -26,7 +26,6 @@ class AllocationActionIT : IndexStateManagementRestTestCase() {
         val states = listOf(
             State("Allocate", listOf(actionConfig), listOf())
         )
-
         val policy = Policy(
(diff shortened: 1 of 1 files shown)


src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/AllocationActionIT.kt
```diff
@@ -100,9 +97,6 @@ class AllocationActionIT : IndexStateManagementRestTestCase() {
             assertEquals(actionConfig.require["_name"], settings["index.routing.allocation.require._name"])
         }
 
-        // Third execution: Waits for allocation to complete, which will happen in this execution since index is small
```
bowenlan-amzn, line 103: No need to trigger the third execution, since the second execution already apply the allocation setting, and we just need to wait for the setting takes effect

The third execution trigger seems collide with the deletion of ISM system index in multi-node test, which could like to a corrupt node lock. Check the analysis here https://github.com/opensearch-project/index-management/issues/90#issuecomment-1747627668

AWSHurneyt (approved):

jowg-amazon (approved):

Issue #710, opened 2023-03-06
[BUG] Rollups - SearchRequest missing cancelAfterTimeInterval parameter


petardz:
**What is the bug?**
In RollupRunner we don't set cancelAfterTimeInterval parameter on SearchRequest object. Consequence of this is that global cluster setting **_search.cancel_after_time_interval_**  will override it, which if set too low, could impact rollup run success rate.

**How can one reproduce the bug?**
1. Set cluster setting **_search.cancel_after_time_interval_** to something very low, like 1s
2. Create rollup job on index with some number of documents/shards
3. See Cancellation timeout error (TaskCancelledExceptioncancelled exception)

**What is the expected behavior?**
Rollup job should have full control of this timeout.

rishabhmaurya:
Are you proposing its resolution to be individual plugin's responsibility to set the `search.cancel_after_time_interval` with `max(cluster setting search.cancel_after_time_interval, plugin's default minimum value say 5s)` ?

petardz:
@rishabhmaurya If I understood correctly, cluster `setting search.cancel_after_time_interval` will be used [only if same param in SearchRequest is null](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/TimeoutTaskCancellationUtility.java#L57)?

If that's the case, then plugin can override this setting but setting a value in every SearchRequest before calling _search

rishabhmaurya:
relates to - https://github.com/opensearch-project/alerting/issues/827

Joshua152:
Can I be assigned to this issue?

PR #1026, opened 2023-10-31
Added minimum for search.cancel_after_time_interval setting for rollups

Joshua152:
*Issue 710, if available:*
#710 

*Description of changes:*
Added line to make the minimum value for cancel_after_time_interval during rollups to 10 minutes

*CheckList:*


src/main/kotlin/org/opensearch/indexmanagement/rollup/RollupSearchService.kt
```diff
@@ -103,7 +111,12 @@ class RollupSearchService(
                             "Composite search failed for rollup, retrying [#${retryCount - 1}] -" +
                                 " reducing page size of composite aggregation from ${job.pageSize} to $pageSize"
                         )
-                        search(job.copy(pageSize = pageSize).getRollupSearchRequest(metadata), listener)
+
+                        val searchRequest = job.copy(pageSize = pageSize).getRollupSearchRequest(metadata)
+                        val timeoutMins = max(cancelAfterTimeInterval.minutes(), 10)
```
bowenlan-amzn, line 116: Can you extract this to be a contant with a good name

bowenlan-amzn:
@Joshua152 this looks good, do you want to create another PR to do the same thing for transform

src/main/kotlin/org/opensearch/indexmanagement/rollup/RollupSearchService.kt
```diff
@@ -103,7 +112,12 @@ class RollupSearchService(
                             "Composite search failed for rollup, retrying [#${retryCount - 1}] -" +
                                 " reducing page size of composite aggregation from ${job.pageSize} to $pageSize"
                         )
-                        search(job.copy(pageSize = pageSize).getRollupSearchRequest(metadata), listener)
+
+                        val searchRequest = job.copy(pageSize = pageSize).getRollupSearchRequest(metadata)
+                        val timeoutMins = max(cancelAfterTimeInterval.minutes(), MINIMUM_CANCEL_AFTER_TIME_INTERVAL_MINUTES)
```
bowenlan-amzn, line 117: Just realise there's a awkward scenario
by default this cancelAfterTimeInterval is -1, [doc](https://opensearch.org/docs/latest/api-reference/search/)
So here we are hard code the cancelAfterTimeInterval for rollup to be 10m, I think it's better to check if the setting is -1 then just don't set anything.
Can extract this out to be a method, and comment why we add this. And easier to add a unit test for this method.

src/main/kotlin/org/opensearch/indexmanagement/rollup/settings/RollupSettings.kt
```diff
@@ -20,6 +20,7 @@ class RollupSettings {
         const val DEFAULT_RENEW_LOCK_RETRY_DELAY = 1000L
         const val DEFAULT_CLIENT_REQUEST_RETRY_COUNT = 3
         const val DEFAULT_CLIENT_REQUEST_RETRY_DELAY = 1000L
+        const val MINIMUM_CANCEL_AFTER_TIME_INTERVAL_MINUTES = 10L
```
Hailong-am, line 23: could this be a ISM specific cluster settings and default value is 10 minutes? @bowenlan-amzn WDYT?

bowenlan-amzn: Yes, it will be more flexible. @Joshua152 can do this as a follow up.

bowenlan-amzn (approved):

jowg-amazon (approved):

PR #1036, opened 2023-11-08
Interval schedule should take start time from the request, should not…

ikibo:
… set it to the current time of request execution.

*Issue #801*

*Description of changes:*
If the transform job schedule is interval, the start job time should be taken from the request
before it was always set to the current time.

*CheckList:*


bowenlan-amzn (approved):
Thanks!

Hailong-am (approved):

jowg-amazon (approved):

Issue #656, opened 2023-01-10
[FEATURE] Allow Transform job to target alias index


dragosrosculete:
Right now Transform can only target an existing index or create a new one. It does not work if an alias index is provided.
This makes it impossible to automate the process and integrate with rollover jobs .

More details:
Lets take the following indexes as example, created hourly

elb-metrics-2022-12-19-01; 
elb-metrics-2022-12-19-02; 
elb-metrics-2022-12-19-03 ; etc ..

From transform job we can select as a source a wildcard of indexes, for example elb-metrics-*  ( this will include present and future indexes). But on the target index I only have the option to select an existing index, for example transformed-metrics or it will create this one if it doesn’t already exist . 
Now this transform jobs runs continuously, so once it is set up, it will write from source (wildcard indexes) to this transformed-metrics index.
The problem is that this index transformed-metrics will eventually have months to years of data but we only want to keep it for a few months.
And there is no way to rotate or create transformed-metrics-2 because the continuous job is set to write into transformed-metrics.
I want to be able to write in a target alias if possible so I can then rotate these indexes and keep only some.

I tried to have Transform to a write into an alias index but this doesn’t work, it does not support to target an alias 
Error: Failed to index the documents

bowenlan-amzn:
Similar request for rollup 
#61 
#445

bowenlan-amzn:
#226 is related and best to be resolved along with this one

n-dohrmann:
Could I be assigned to this?

MahendraAkkina:
+ 1. This is a huge limitation for us to use transform jobs.

MahendraAkkina:
@dragosrosculete How were you able to get around this problem?

franco-martin:
@MahendraAkkina I used an index pipeline to rewrite the _index field to the alias. Worked like a charm. It doesn't work with datastreams because of the op_type.

MahendraAkkina:
@franco-martin Can you provide an example? Thanks

PR #1049, opened 2023-11-29
Alias index transform

n-dohrmann:
*Issue #, if available:*
#656 

*Description of changes:*
Adding to ``createTargetIndex`` method in the ``TransformIndexer`` module - want to support using aliases for variable target indices on transform jobs.

*CheckList:*
- [ x] Commits are signed per the DCO using --signoff


src/main/kotlin/org/opensearch/indexmanagement/transform/TransformIndexer.kt
```diff
@@ -63,6 +65,20 @@ class TransformIndexer(
...
+                throw TransformIndexException("Target index alias has no write index")
+            }
+        }
+        val putMappingReq = PutMappingRequest(writeIndexMetadata?.index?.name).source(targetFieldMappings)
+        val mapResp: AcknowledgedResponse = client.admin().indices().suspendUntil {
+            putMapping(putMappingReq)
+        }
+        if (!mapResp.isAcknowledged) {
+            logger.error("Target index mapping request failed")
+        }
```
bowenlan-amzn, line 81: Also move this inside to above if block?

HenryTheSir:
Hi, 
is there any news on this feature?
Best regards

MahendraAkkina:
Any update on this?

bowenlan-amzn:
@HenryTheSir @MahendraAkkina I don't think @n-dohrmann still works on this. Would you want to create another PR from this?

MahendraAkkina:
@bowenlan-amzn Thanks for checking on this. I am not familiar with the dev process. Is there a doc talking about it?. I see that you were involved heavily in reviewing it. So wanted to check if its something you can help with creating a new PR and help with getting it merged?
cc: @HenryTheSir

bowenlan-amzn:
@MahendraAkkina this doc should help https://github.com/opensearch-project/index-management/blob/main/DEVELOPER_GUIDE.md
I'm mainly working on another projects now and limiting here, sorry about that. 
Let me tag current active maintainer to see if he can help @vaibhoag

MahendraAkkina:
@bowenlan-amzn @vaibhoag I am looking at the fix and am little confused. I see that createTargetIndex is failing as it is not able to create an index which is actually an alias. The code fix has additional code after the first if. I think all this code needs to go under first if check. @bowenlan-amzn Is that what you were suggesting?

MahendraAkkina:
Something like this, that is I moved the additional code into the first if
```
    private suspend fun createTargetIndex(targetIndex: String, targetFieldMappings: Map<String, Any>) {
        if (!clusterService.state().routingTable.hasIndex(targetIndex)) {
            if (clusterService.state().metadata.hasAlias(targetIndex)) {
                // return error if no write index with the alias
                val writeIndexMetadata = clusterService.state().metadata.indicesLookup[targetIndex]!!.writeIndex
                if (writeIndexMetadata == null) {
                    throw TransformIndexException("target_index [$targetIndex] is an alias but doesn't have write index")
                }
                val putMappingReq = PutMappingRequest(writeIndexMetadata.index?.name).source(targetFieldMappings)
                val mapResp: AcknowledgedResponse = client.admin().indices().suspendUntil {
                    putMapping(putMappingReq)
                }
                if (!mapResp.isAcknowledged) {
                    logger.error("Target index mapping request failed")
                }
            } else {
                val transformTargetIndexMapping = TargetIndexMappingService.createTargetIndexMapping(targetFieldMappings)
                val request = CreateIndexRequest(targetIndex).mapping(transformTargetIndexMapping)
                // TODO: Read in the actual mappings from the source index and use that
                val response: CreateIndexResponse = client.admin().indices().suspendUntil { create(request, it) }
                if (!response.isAcknowledged) {
                    logger.error("Failed to create the target index $targetIndex")
                    throw TransformIndexException("Failed to create the target index")
                }
            }
        }
    }
```

bowenlan-amzn:
[**@MahendraAkkina**](https://github.com/MahendraAkkina) commented on [Apr 16, 2025, 6:00 PM PDT](https://github.com/opensearch-project/index-management/pull/1049#issuecomment-2811375436 "2025-04-17T01:00:56Z - Replied by Github Reply Comments"):
> I am looking at the fix and am little confused. I see that createTargetIndex is failing as it is not able to create an index which is actually an alias. The code fix has additional code after the first if. I think all this code needs to go under first if check. Is that what you were suggesting?

Yeah, we should move the alias check to the top, first see if provided targetIndex is an alias, then see if we want to create that index.

MahendraAkkina:
Instead of updating the mappings every time a transform runs, especially in continuous mode, I think choosing not to update them for aliases. My reasoning is that updating mappings on each run feels inconsistent with how index mappings are handled, where they are typically defined once at index creation. In the specific scenario with rollover aliases, we should rely on index templates to pre-define all the necessary fields when new indices are created, thus eliminating the need for continuous mapping updates.

@vaibhoag, thoughts? Can you or someone else at amazon :-) help with contributing to the this project?
cc: @bowenlan-amzn 
```
    private suspend fun createTargetIndex(targetIndex: String, targetFieldMappings: Map<String, Any>) {
        if (!clusterService.state().routingTable.hasIndex(targetIndex)) {
            if (clusterService.state().metadata.hasAlias(targetIndex)) {
                logger.debug("Checking to see if target_index [$targetIndex] that is an alias has write index")
                // return error if no write index with the alias
                val writeIndexMetadata = clusterService.state().metadata.indicesLookup[targetIndex]!!.writeIndex
                if (writeIndexMetadata == null) {
                    throw TransformIndexException("target_index [$targetIndex] is an alias but doesn't have write index")
                }
            } else {
                val transformTargetIndexMapping = TargetIndexMappingService.createTargetIndexMapping(targetFieldMappings)
                val request = CreateIndexRequest(targetIndex).mapping(transformTargetIndexMapping)
                // TODO: Read in the actual mappings from the source index and use that
                val response: CreateIndexResponse = client.admin().indices().suspendUntil { create(request, it) }
                if (!response.isAcknowledged) {
                    logger.error("Failed to create the target index $targetIndex")
                    throw TransformIndexException("Failed to create the target index")
                }
            }
        }
    }
```

vaibhoag:
Hey @MahendraAkkina I think instead of updating mappings, we should first validate the `writeIndex` mapping with `targetFieldMappings` and if that is false, then we should update the mappings

Actually this depends on how we are creating targetAlias, are we expecting users to create it? @bowenlan-amzn

MahendraAkkina:
@vaibhoag That should work. I would assume it would check this every time the transform operation runs (Example: In the case when continuous is set to true and is scheduled to run every hour, this check would happen once an hour). We can imagine the writeIndex to be the same for a long time (as it typically depends on the rollover policy). Might not be a big performance hit, but just wanted to point out.
Alternatively, I am ok with having a limitation for alias case where we expect the mappings to already exist in the writeIndex. Rationale being, we are not creating the index in this case.

MahendraAkkina:
@vaibhoag Will you be able to help with this getting into a release. The code provided above should work.

Issue #1055, opened 2023-12-10
[BUG] Listing snapshot-management policies fails when the ism index is not yet created


aggarwalShivani:
**What is the bug?**
Index-management plugin is installed in Opensearch. No policies are created yet, so the index .opendistro-ism-config does not exist yet.

Listing ism policies does not fail, and prints an empty list.
```
$ curl 127.0.0.1:9200/_plugins/_ism/policies
{"policies":[],"total_policies":0}
```

However, listing snapshot-management policies throws an exception due to absence of the index.
```
curl 127.0.0.1:9200/_plugins/_sm/policies
{"error":{"root_cause":[{"type":"status_exception","reason":"Snapshot management config index not found"}],"type":"status_exception","reason":"Snapshot management config index not found"},"status":404}
```

**How can one reproduce the bug?**
1. Install Opensearch with index-management plugin
2. After the cluster is ready (healthy), run the above mentioned GET cmds to list ism and sm policies.
3. See error while listing sm policies.

**What is the expected behavior?**
Like ism policies, even the sm api could respond with an empty list, instead of an exception.

**What is your host/environment?**
 - OS: Windows 10
 - Version 2.8 (latest branch)
 - Plugins

aggarwalShivani:
Hi
The expectation is consistent behaviour between ism and sm policies. Either both APIs must throw similar exception when the index does not exist, or both should return an empty list.
Request feedback from repo maitainers to decide next steps for this issue,

aggarwalShivani:
Hi, 
Request feedback from maintainers on whether the requested change is valid/desirable from the plugin's design, and if yes, I would be interested to pick this item to contribute.

aggarwalShivani:
Hi, 
Gentle reminder: Request feedback from maintainers on this issue. 
If this change is desirable, I would be interested to start working on the PR. I have already identified a potential fix.

aggarwalShivani:
Hi, I have raised a PR for this change. Request maintainers for review and feedback.

aggarwalShivani:
Issue fixed with https://github.com/opensearch-project/index-management/pull/1072

PR #1072, opened 2024-01-16
Fix for GET request on SM policies to return empty list when ism config index does not exist

aggarwalShivani:
Fix for GET request on SM policies to return empty list when ism config index does not exist

*Issue #, if available:* #1055

*Description of changes:*
When a GET request is issued to snapshot mananagement API endpoint, to list all available sm policies, it throws an IndexNotFound exception if the ism config index is not yet created. (Note - This index gets created only after atleast one ism/sm policy gets created).
To align the behaviour to ism policies endpoint (that responds with an empty list for missing config index cases), the change is done here for SM policies.
In success case, searchResponse is returned like done before. 
In the failure case (for missing config index), instead of throwing OpenSearchStatusException, an empty list of 0 Length is returned from the getAllPolicies function.
If any other exception is caught, then that exception is thrown as-is. 

*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/snapshotmanagement/api/transport/get/TransportGetSMPoliciesAction.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/snapshotmanagement/api/transport/get/TransportGetSMPoliciesAction.kt
@@ -70,11 +70,15 @@ class TransportGetSMPoliciesAction @Inject constructor(
     private suspend fun getAllPolicies(searchParams: SearchParams, user: User?): Pair<List<SMPolicy>, Long> {
         val searchRequest = getAllPoliciesRequest(searchParams, user)
         val searchResponse: SearchResponse = try {
-            client.suspendUntil { search(searchRequest, it) }
+        return try {
+            val searchResponse = client.suspendUntil { search(searchRequest, it) }
+            parseGetAllPoliciesResponse(searchResponse)
         } catch (e: IndexNotFoundException) {
-            throw OpenSearchStatusException("Snapshot management config index not found", RestStatus.NOT_FOUND)
+            // config index hasn't been initialized, catch this here and show empty result for policies
+            Pair(emptyList(), 0L)
+        } catch (e: Exception) {
+            throw e
         }
-        return parseGetAllPoliciesResponse(searchResponse)
     }
 
     private fun getAllPoliciesRequest(searchParams: SearchParams, user: User?): SearchRequest {


src/main/kotlin/org/opensearch/indexmanagement/snapshotmanagement/api/transport/get/TransportGetSMPoliciesAction.kt
```diff
@@ -69,12 +69,15 @@ class TransportGetSMPoliciesAction @Inject constructor(
 
     private suspend fun getAllPolicies(searchParams: SearchParams, user: User?): Pair<List<SMPolicy>, Long> {
         val searchRequest = getAllPoliciesRequest(searchParams, user)
-        val searchResponse: SearchResponse = try {
-            client.suspendUntil { search(searchRequest, it) }
+        return try {
+            val searchResponse = client.suspendUntil { search(searchRequest, it) }
+            parseGetAllPoliciesResponse(searchResponse)
         } catch (e: IndexNotFoundException) {
```
Hailong-am, line 75: would like to `ExceptionHelper.unwrapCause(e)` for multiple node case, and please add test for the change.

aggarwalShivani: Thanks for the feedback @Hailong-am. 
I have handled the exception with ExceptionHelper.unwrapCause(e) in the final catch. I hope my understanding was right. Please let me know if there is still any additions needed. 
I have also added a test-case for this change.

Hailong-am: sorry for the misunderstanding, i was thinking add the unwrap to line 76 when checking `IndexNotFound` exception.

Hailong-am:
@bowenlan-amzn would like to confirm with you, return empty list meet your expectation as well

bowenlan-amzn:
@Hailong-am Yes, this is the behavior of ISM. Though rollup and transform doesn't catch IndexNotFoundException

src/main/kotlin/org/opensearch/indexmanagement/snapshotmanagement/api/transport/get/TransportGetSMPoliciesAction.kt
aggarwalShivani: okay sure, but for IndexNotFound case, this change would return empty list and not throw the exception. So I wonder where to add the unwrap here.

Instead, we could combine the two catch blocks something like below. Does this look okay?

```
try {
 ....
} catch (e: Exception) {
    val unwrappedException = ExceptionsHelper.unwrapCause(e) as Exception
    if (unwrappedException is IndexNotFoundException) {
	// config index hasn't been initialized, catch this here and show empty result for policies
        Pair(emptyList(), 0L)
    } else {
        throw unwrappedException
    }
}
```

aggarwalShivani:
Hi reviewers, 
Thanks for the feedback so far. 
I'm facing difficulty in understanding some failures in checks running on the github CI for my PR. Its my first PR to this project and would highly appreciate any pointers and guidance. 
4 out of 30 checks have failed. I've put some details below -

i. [Docker Security Test Workflow](https://github.com/opensearch-project/index-management/actions/runs/7571807072/job/20668374157?pr=1072)
```
WARNING: Please consider reporting this to the maintainers of org.opensearch.bootstrap.Security
uncaught exception in thread [main]
java.lang.IllegalStateException: failed to load plugin class [org.opensearch.security.OpenSearchSecurityPlugin]
Likely root cause: OpenSearchException[plugins.security.ssl.transport.keystore_filepath or plugins.security.ssl.transport.server.pemcert_filepath and plugins.security.ssl.transport.client.pemcert_filepath must be set if transport ssl is requested.]
```

Analysis - This looks like an issue with security plugin, unrelated to the changes made in this PR. Any thoughts on this?


ii. In both these cases,  [Test and Build Workflow / test-and-build-linux (21, ism)](https://github.com/opensearch-project/index-management/actions/runs/7571807134/job/20668382458?pr=1072) and [Test and Build Workflow / test-and-build-linux (21, non-ism)](https://github.com/opensearch-project/index-management/actions/runs/7571807134/job/20668382665?pr=1072), all failures have the same exception ->
```
   Caused by:
        java.net.SocketTimeoutException: 60000 MILLISECONDS
java.lang.AssertionError: Fri Jan 19 15:23:33 ART 2024: There are still tasks running after this test that might break subsequent tests:
```

   I have run some of these testcases individually on my local setup, and they passed. Any pointers on the root cause here? Are they expected?
   
   
iii. [codecov/patch](https://github.com/opensearch-project/index-management/pull/1072/checks?check_run_id=20668810974) failed with warning - Added lines #L78 - L79 were not covered by tests.
But the test-case added, does cover the mentioned case. I'll perhaps recheck this after changes get finalized for [TransportGetSMPoliciesAction.kt](https://github.com/opensearch-project/index-management/pull/1072/files/2ac032a332f1d9dce0a140878ffbe3f97ab65879#diff-17c2b890310e087e0f6489f854b3d793a12f06b97ee0d5f0501cabbe04e45445)

src/main/kotlin/org/opensearch/indexmanagement/snapshotmanagement/api/transport/get/TransportGetSMPoliciesAction.kt
Hailong-am: > okay sure, but for IndexNotFound case, this change would return empty list and not throw the exception. So I wonder where to add the unwrap here.
> 
> Instead, we could combine the two catch blocks something like below. Does this look okay?
> 
> ```
> try {
>  ....
> } catch (e: Exception) {
>     val unwrappedException = ExceptionsHelper.unwrapCause(e) as Exception
>     if (unwrappedException is IndexNotFoundException) {
> 	// config index hasn't been initialized, catch this here and show empty result for policies
>         Pair(emptyList(), 0L)
>     } else {
>         throw unwrappedException
>     }
> }
> ```

Yes, that's correct, thanks.

Hailong-am:
> ii. In both these cases, [Test and Build Workflow / test-and-build-linux (21, ism)](https://github.com/opensearch-project/index-management/actions/runs/7571807134/job/20668382458?pr=1072) and [Test and Build Workflow / test-and-build-linux (21, non-ism)](https://github.com/opensearch-project/index-management/actions/runs/7571807134/job/20668382665?pr=1072), all failures have the same exception ->

we can ignore jdk 21 failure, it's known issue. Reference https://github.com/opensearch-project/flow-framework/issues/426 

> [Docker Security Test Workflow](https://github.com/opensearch-project/index-management/actions/runs/7571807072/job/20668374157?pr=1072)

Related to https://github.com/opensearch-project/index-management/issues/1064, will have another PR to handle it.

Hailong-am (approved):

bowenlan-amzn (approved):

aggarwalShivani:
Thanks @Hailong-am and @bowenlan-amzn for the inputs and help here :)

Issue #1127, opened 2024-03-08
[FEATURE] Opt to run ISM jobs even on red cluster state


aggarwalShivani:
**Is your feature request related to a problem?**

As mentioned in [ISM docs](https://opensearch.org/docs/latest/im-plugin/ism/index/#step-2-attach-policies-to-indexes) and observed in my experiments, 

> **ISM does not run jobs if the cluster state is red.**

We use ISM policies (with delete action) to regularly delete old indices from the cluster and free up disk space.
Inspite of regular cleanup, if there is high data ingestion, sometimes the cluster could go to red state due to low disk availability. In such cases, the problem is, even though we have setup ISM policies, they would not run and the cluster would not get recovered, unless someone manually runs the DELETE REST API to cleanup indices and recovering the cluster-state.

**What solution would you like?**
There could be a new optional parameter (for ex. allow_red_cluster) for all actions or atleast the delete action - that a user could set, if they wish to run the specified ISM policies even on a red cluster.

mgodwan:
@aggarwalShivani I think this sounds like a fair enough use case to allow within ISM policy executions. Would you be willing to raise a PR for this?

skumarp7:
Hi @mgodwan, 

can you assign this to me ? I would like to work on this :)

mgodwan:
Awesome @skumarp7 . Assigned to you.

skumarp7:
Hi @mgodwan , 

I was trying to add a new optional parameter ( ex: allow_red_cluster ] in the delete operation specifically under the action as few operations such as shrink, force_merge might not make sense to run on a red cluster. But this is currently a problem as if there are more states, there will not be any transitions and hence it won't reach to the state where the delete operation is defined and even though the parameter is set to true.  

Should we add the optional parameter at the top level of the policy or do we have any another suggestion from you ?

aggarwalShivani:
Hi @mgodwan and other maintainers,
Any suggestions on the query asked by @skumarp7 ?

In my view, it may be misleading to add the "allow_red_cluster" parameter at the policy level if all actions do not / may not support running on a red-cluster. 

One alternative approach could be adding this parameter at the [transitions](https://opensearch.org/docs/latest/im-plugin/ism/policies/#transitions) level in the policy. For ex.
```
"transitions": [
  {
    "state_name": "cold",
    "allow_red_cluster": true,
    "conditions": {
      "min_index_age": "30d"
    }
  }
]
```

[...]

PR #1189, opened 2024-06-13
Addition of Optional Parameter - allow_Red_cluster to bypass ISM to run on a Red Opensearch Cluster

skumarp7:
*Issue #, if available:*

https://github.com/opensearch-project/index-management/issues/1127

*Description of changes:*

This PR involves the provision for ISM to run on a red opensearch cluster. 

**Behaviour before:** 

- The current implementation of the managedIndexRunner job prohibits any calls to the cluster when the cluster is red. 
  https://github.com/opensearch-project/index-management/blob/a049fdb423efcb02fb08a10f3b984fe392d09bab/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt#L254
  
- In case if the cluster goes to red state due to resource crunch, ISM cannot delete any indices due to the above constraint. 

- Consider each index is of size 1 GB and the persistent volume of the opensearch cluster is 10GB. 

_ISM on a Green/Yellow Opensearch Cluster:_ 

ISM

- When the cluster volume reaches 95% of the available volume, the cluster goes to red state. 

_ISM on a Red Opensearch Cluster:_

ISM-Red

**Behaviour Post the Changes:** 

- The user has the option to set "allow_red_cluster" for each state where all the operations on that state can run on a red cluster if the "allow_red_cluster" is set to true. This parameter is optional. 

- If the User sets the parameter to "true", the ISM job will be bypassed to run on the particular state on a red Cluster. Any transitions, Actions under that state will be executed. 

- By default, the value of "allow_red_cluster" is false. Hence, the current behaviour of the API is also unchanged if the user doesnt want to bypass ISM to be run on a red cluster. 

_ISM on a Green/Yellow Opensearch Cluster:_ 

ISM-GY-WITH PARAM
****

_ISM on a Red Opensearch Cluster with the optional parameter:_

ISM - 1

- In case if the cluster goes to red state due to the volume getting consumed more than 95%, the ISM will still execute the delete action because of the param allowRedCluster being set true, which will delete the index which is of size 1 GB. This would help us to get the cluster back to yellow/green state and the ISM will continue to run successfully for the previous stages. 
 
*CheckList:*

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
@@ -250,11 +250,6 @@ object ManagedIndexRunner :
     @Suppress("ReturnCount", "ComplexMethod", "LongMethod", "ComplexCondition", "NestedBlockDepth")
     private suspend fun runManagedIndexConfig(managedIndexConfig: ManagedIndexConfig, jobContext: JobExecutionContext) {
         logger.debug("Run job for index ${managedIndexConfig.index}")
-        // doing a check of local cluster health as we do not want to overload cluster manager node with potentially a lot of calls
-        if (clusterIsRed()) {
-            logger.debug("Skipping current execution of ${managedIndexConfig.index} because of red cluster health")
-            return
-        }
 
         val (managedIndexMetaData, getMetadataSuccess) = client.getManagedIndexMetadata(managedIndexConfig.indexUuid)
         if (!getMetadataSuccess) {
@@ -314,6 +309,15 @@ object ManagedIndexRunner :
 
         val state = policy.getStateToExecute(managedIndexMetaData)
         val action: Action? = state?.getActionToExecute(managedIndexMetaData, indexMetadataProvider)
+        val allowRedCluster = state?.allowRedCluster
+        // doing a check of local cluster health as we do not want to overload cluster manager node with potentially a lot of calls
+        if (clusterIsRed()) {
+            if (allowRedCluster == false) {
+                logger.debug("Skipping current execution of ${managedIndexConfig.index} because of red cluster health")
+                return
+            }
+            logger.warn("Warning: ISM is running on a red cluster")
+        }
         val stepContext =
             StepContext(
                 managedIndexMetaData, clusterService, client, threadPool.threadContext, policy.user, scriptService, settings, jobContext.lockService,
--- a/src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/runner/ManagedIndexRunnerIT.kt
+++ b/src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/runner/ManagedIndexRunnerIT.kt
@@ -5,6 +5,11 @@
 
 package org.opensearch.indexmanagement.indexstatemanagement.runner
 
+import org.apache.hc.core5.http.ContentType
+import org.apache.hc.core5.http.io.entity.StringEntity
+import org.opensearch.client.Request
+import org.opensearch.client.Response
+import org.opensearch.core.rest.RestStatus
 import org.opensearch.indexmanagement.indexstatemanagement.ISMActionsParser
 import org.opensearch.indexmanagement.indexstatemanagement.IndexStateManagementRestTestCase
 import org.opensearch.indexmanagement.indexstatemanagement.action.OpenAction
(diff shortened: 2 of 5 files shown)


skumarp7:
Hi ,

Request feedback from the maintainers :)

skumarp7:
Hi All,

I was trying to write test for the above feature and facing issues to simulate a red cluster. I tried to bring down the data node to make the cluster go read but since the integTest clusters are immutable, that was not possible. 

Is there any other approach that i can make use of to simulate the red cluster ?

Hailong-am:
> Hi All,
> 
> I was trying to write test for the above feature and facing issues to simulate a red cluster. I tried to bring down the data node to make the cluster go read but since the integTest clusters are immutable, that was not possible.
> 
> Is there any other approach that i can make use of to simulate the red cluster ?


you can specify allocation policy for a index, An example
```
PUT test
{
  "settings": {
    "index.routing.allocation.require.size": "big"
  }
}
```
it will create index `test` and allocate shards on node with attribute size is big. As we don't have such node, the primary shards will not able to allocate, that will cause cluster health is red.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
```diff
@@ -250,11 +250,6 @@ object ManagedIndexRunner :
     @Suppress("ReturnCount", "ComplexMethod", "LongMethod", "ComplexCondition", "NestedBlockDepth")
     private suspend fun runManagedIndexConfig(managedIndexConfig: ManagedIndexConfig, jobContext: JobExecutionContext) {
         logger.debug("Run job for index ${managedIndexConfig.index}")
-        // doing a check of local cluster health as we do not want to overload cluster manager node with potentially a lot of calls
```
bowenlan-amzn, line 253: We shouldn't move this  red cluster check down. Because it helps to not making 1 or more calls below which potentially worsen a red cluster. 

Instead, we would need to rely on the ManagedIndexConfig which we directly have here. It has a snapshot of the policy. Only when the policy has `allow_red_cluster`, then we will bypass this check.

The problem is ManagedIndexConfig doesn't know which state or action is running, only metadata knows, so we won't be able to just bypass when specific state or action is running. To do that, more complex logic would be needed to update ManagedIndexConfig with required info...

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/runner/ManagedIndexRunnerIT.kt
```diff
@@ -224,4 +229,101 @@ class ManagedIndexRunnerIT : IndexStateManagementRestTestCase() {
...
+            "actions":[],"transitions":[{"state_name":"inactivestate","conditions":{"min_index_age":"5s"}}]},{"name":"inactivestate","allow_red_cluster":"true"
+            ,"actions":[{"delete":{}}],"transitions":[]}],"ism_template":{"index_patterns":["test-index"]}}}
+            """.trimIndent()
+        createPolicyJson(policy, policyID)
+        createIndex(indexName, policyID)
+        waitFor { assertIndexExists(indexName) }
+
+        val managedIndexConfig = getExistingManagedIndexConfig(indexName)
+
+        val endpoint = "sim-red"
```
bowenlan-amzn, line 296: Would it make sense to not use a different index, but just use `val indexName = "test-index-2"` to simulate the red cluster situation?

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
skumarp7: Hi @bowenlan-amzn, 

Hi @bowenlan-amzn, 

The calls we're making to the cluster are mostly GET operations, to fetch state etc, so I'm not sure how they would worsen a red cluster. I can still try to avoid them, but not sure of the right approach --
 
This is my understanding of your suggestions. Please correct where required :
- Move the check to original location
- Do some parsing of the policy to figure out if "allow_red_cluster" is present anywhere in the policy. If it does not, return like we do in existing code.
- Else allow to run further 
(At this stage, we don't know for which state this is going to be executed, as we cannot run any get calls).
 
 
Challenges:
- allow_red_cluster flag has been added at state level i.e. some states in a policy can have it enabled true and some might not.
- After we have allowed it to bypass the initial policy check, we again require a check at the state/action level if parameter is set to 'true or false' to  run/skip that state/action.
 
That means, for a state that didn't have allow_red_cluster true, we would still execute those 1-2 GET calls before skipping it - which again could potentially worsen the cluster.
 
 
Ques:
1. How would be able to get the value of "allow_red_cluster" parameter at each state before even we get that metadata? can you throw some light to help me implement this logic in ManagedIndexConfig?
2. If we still need to place the check at top level only, we can move the parameter at the policy level itself instead of state level (as that can be fetched from the ManagedIndexConfig). 
That would mean run all state/actions (if flag is true), or run nothing (if flag is false).
But with this, the concern rises that - any ism action that doesnt support red cluster or harm a red cluster will also be bypassed to run on a red cluster.

bowenlan-amzn: > The calls we're making to the cluster are mostly GET operations, to fetch state etc, so I'm not sure how they would worsen a red cluster.

Right, these are get calls, but consider this small change will affect all the ISM jobs which can be 10k+ in a cluster, we should still keep the original logic — don't do anything if cluster is red.

> That would mean run all state/actions (if flag is true), or run nothing (if flag is false).
But with this, the concern rises that - any ism action that doesnt support red cluster or harm a red cluster will also be bypassed to run on a red cluster.

We don't need new logic in ManagedIndexConfig. ManagedIndexConfig has policy in it
https://github.com/opensearch-project/index-management/blob/dbd2bc25f5e338819237af5a6b35c6c1e35d7171/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/ManagedIndexConfig.kt#L38

I'm thinking a little hacky solution maybe having sth like `allow_red_cluster_after: 3d`, then compare the current time with `jobEnabledTime` which seems to be the time when the job is created. So we can calculate when to start to allow red cluster, before then, we still block execution by red cluster.

skumarp7: Hi @bowenlan-amzn,
 
I am guessing the idea here being suggested is we give a time-window to the user to decide - such that if the cluster continues to remain in red state for that long, then ism would be allowed to run. 
 
Few follow up questions:
- Let's say that the policy has triggered the job on day 1 and the cluster state goes to red after 10 days. The job would immediately get triggered on the red-cluster as the difference between the `jobEnabledTime`  and current time would already have exceeded 3 days. In this case, the red cluster will be bypassed regardless of the time duration of 3days set in the parameter and it would not wait in red state for 3d.  Is this understanding correct?

- Is there any way to retrieve information for how many days the cluster state has been in Red?
Just thinking out loud on whether we could instead check the duration of how much time the cluster state has been red and compare that with `allow_red_cluster_after: 3d`.

vikasvb90:
@skumarp7
 Instead of allowing all ISM actions to bypass red cluster, can we enable it specifically on actions which potentially move towards reducing cluster redness? For instance, delete index moves towards reducing red cluster.
For some actions to not worsen the cluster, we probably need scoping and some like create index, rollover, etc will always add to red state. So, lets even not add any means to allow these actions in case of red cluster state. 
There are cases where shard assignments are blocked because of some specific reasons, but handling those specific scenarios and evaluating and allowing actions may be quite complex to achieve.

To summarize, I am aligned to allowing executions of ISM actions which can either improve cluster state or are neutral but definitely not on actions which can make it worse.

vikasvb90 (commented):
@skumarp7
Instead of allowing all ISM actions to bypass red cluster, can we enable it specifically on actions which potentially move towards reducing cluster redness? For instance, delete index moves towards reducing red cluster.
For some actions to not worsen the cluster, we probably need scoping and some like create index, rollover, etc will always add to red state. So, lets even not add any means to allow these actions in case of red cluster state.
There are cases where shard assignments are blocked because of some specific reasons, but handling those specific scenarios and evaluating and allowing actions may be quite complex to achieve.

To summarize, I am aligned to allowing executions of ISM actions which can either improve cluster state or are neutral but definitely not on actions which can make it worse.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ManagedIndexRunner.kt
aggarwalShivani: Hi @bowenlan-amzn 
I checked a bit on this, sharing my observations on this.
jobEnabledTime is the time when the ism job for the index gets created. It does not get updated as per the running state/action and remains the same until the job gets completed for an index.

```
Lets say the policy is created that deletes indices after they reach age > 5d and allow_red_cluster_after: 3d is also set.
On day 1, job starts and jobEnabledTime gets set to this time.
On day 4, cluster goes to red.
On day 5, it has already exceeded 3d from jobEnabledTime and the red cluster check would get bypassed. 
```

So if the intention was to make the job wait in the red state for configured time (3d), comparison with jobEnabledTime is not working as expected.
We could however make use of **action startTime** from the metadata. 
For ex. On day 5, delete action would get started, but it would get skipped as cluster is in red state. This action's start time could be checked with current time, and after this exceeds 3d, this action could be allowed further.

Any thoughts on this...?

aggarwalShivani: Hi @bowenlan-amzn 
Following up on my previous comment, I have tried this approach ->

```
-- Job starts --
If cluster health is red
  if allow_cluster_health_after is enabled and  (current time - actionstarttime) > allow_red_cluster_after:
      proceed
  else
    return, skip execution. 

At state execution:  
  if cluster is red   
     if action is delete
        proceed
     else
       return, skip execution. 
```

This will basically mean if health is red and the action is stuck, it shall wait for configured duration (3d) in that situation and only then allow further execution if the flag is set.
Also, at the action execution, it will recheck that it allows execution only for delete action.
Code snippets for reference->

````
        val indexMetadata = client.getManagedIndexMetadata(managedIndexConfig.indexUuid)
        val actionStartTime = (indexMetadata.first?.let { getActionStartTime(it) })
       if (clusterIsRed()) { 
          if (allowDeleteOnRedEnabled and Instant.now().isAfter(actionStartTime?.plusSeconds(allowRedClusterAfter.seconds))) {
                logger.info("Allowing to run ism after waiting for more than " + allowRedClusterAfter + " + on red state")
            } else {
                logger.info("Skipping current execution of ${managedIndexConfig.index} because of red cluster health")
                return
            }
      }
      .....
      .....   // At the state execution phase 
        val action: Action? = state?.getActionToExecute(managedIndexMetaData, indexMetadataProvider)
        if (clusterIsRed()) {
            // If it has come till here in red-state, it means allow_red_cluster_after duration has passed.
            if (action is DeleteAction) {
                logger.info("Allowing only DeleteAction during red status")
            } else {
                logger.debug("Skipping current execution of ${managedIndexConfig.index} because of red cluster health")
                return
            }
        }
````

With this approach, any other action or transition does not happen on red-state. 
If an index was already transitioned and almost ready to execute the delete action when the cluster became red, only that undergoes delete action, rest of them remain stuck until cluster health gets better.

Please share your thoughts on this..

aggarwalShivani: Hi @bowenlan-amzn 
Any thoughts on my previous comment? 🙂

skumarp7:
closing this PR to open a new one.

PR #1440, opened 2025-07-07
Support for no_alias and min_state_age in ISM TransitionsFeature/ism enhancement

Tarun-kishore:
### Description

This PR introduces two new transition conditions to OpenSearch Index State Management (ISM):

- **`no_alias` (boolean):**
  - Allows transitions to be gated based on whether an index is associated with any aliases.
  - When `true`: transition only if the index has no aliases.
  - When `false`: transition only if the index has at least one alias.

- **`min_state_age` (string duration):**
  - Allows transitions only if the index has spent a minimum duration in the current ISM state.
  - Accepts time values like `"5m"`, `"7d"`.
  - Useful for implementing retention delays or controlled state progression.

#### Why `min_state_age` is required  even though `min_index_age` is supported

The existing `min_index_age` condition measures time since index creation, which is not sufficient for lifecycle transitions that depend on how long an index has been in a specific state.

For example, if an index transitions to an `archive` state several days after creation, using `min_index_age` could cause it to immediately move to the next state — even though it just entered `archive`. Let's say we use a `no_alias` condition to move an index to `archive`; we wouldn't be able to guarantee that it stays in `archive` for a defined period if relying on `min_index_age` alone, since alias removal is independent of index age.

By contrast, `min_state_age` starts counting from when the index enters the current state, allowing policies to enforce precise retention windows or delay transitions. This enables reliable patterns like “stay in archive for 7 days” or “delay deletion for 24 hours after entering a cold state.”

---

### Use Cases

- Prevent deletion or archiving of indexes still in use via aliases.
- Delay transitions (e.g., delete, freeze) for a fixed period after entering a state.
- Improve expressiveness of lifecycle policies to match real-world operational workflows.

---

### Testing

Added integration tests for both new conditions.

Also validated through local testing in a controlled OpenSearch cluster using REST APIs and ISM Explain API.

**Test 1: `no_alias: true`**

- **Setup:** Created an index with an alias and attached a policy with `no_alias: true`.
- **Expected:** Transition should be blocked while alias is present.
- **Action:** Removed alias.
- **Result:** Transition successfully occurred after alias removal.
- **Validation:** Confirmed via ISM Explain API.

---

**Test 2: `no_alias: false`**

- **Setup:** Created a new index without alias, attached a policy with `no_alias: false`.
- **Expected:** Transition should be blocked while alias is absent.
- **Action:** Added alias.
- **Result:** Transition proceeded after alias was added.
- **Validation:** Confirmed via ISM Explain API.

---

**Test 3: `min_state_age`**

- **Setup:** Created an index and attached a policy with `min_state_age: 20m`.
- **Expected:** Transition should be blocked until the index has been in the current state for at least 20 minutes.
- **Action:** Checked status before and after 20-minute window.
- **Result:** Transition was initially blocked, and succeeded after the duration.
- **Validation:** Confirmed via ISM Explain API.

---

### Related Issues
Resolves #[1439]

### Check List


Tarun-kishore:
While ISM provides an `alias` action to programmatically add or remove aliases from an index, the `no_alias` condition plays a fundamentally different role: it **controls when a transition should happen** based on whether the index is currently referenced by any alias.

In simple terms:

- The `alias` **action** tells ISM to remove an alias *before* transitioning.
- The `no_alias` **condition** tells ISM to **wait** and transition **only after all aliases have already been removed**—typically by an external system.

This distinction is essential in workflows like rollovers or index promotions, where alias updates are managed outside ISM (e.g., by ingestion pipelines). Using `no_alias` ensures that lifecycle transitions such as archiving or deletion occur **only after** the index is no longer in active use, preventing premature data removal.

Regarding concerns for having `min_state_age` and `min_index_age` at same time, ISM doesn't allow more than one condition per transition so that should not be a concern.

bowenlan-amzn:
Makes sense. if `no_alias` is to work with external system.

> ISM doesn't allow more than one condition per transition so that should not be a concern.

I see, this feels like a limitation (for example `no_alias` and `min_state_age` can be ANDed together), but helps simplify things here.

Tarun-kishore:
Fixed the CI builds

Tarun-kishore:
Quick question — once this PR is merged, should the [ISM docs](https://docs.opensearch.org/docs/latest/im-plugin/ism/policies/) be updated immediately or only after the next release? Just want to be sure I follow the right process.

bowenlan-amzn:
[**@Tarun Kishore**](https://github.com/Tarun-kishore) commented on [Jul 11, 2025, 8:48 PM PDT](https://github.com/opensearch-project/index-management/pull/1440#issuecomment-3064618675 "2025-07-12T03:48:52Z - Replied by Github Reply Comments"):
> Quick question — once this PR is merged, should the [ISM docs](https://docs.opensearch.org/docs/latest/im-plugin/ism/policies/) be updated immediately or only after the next release? Just want to be sure I follow the right process.

I think we can specify which release in the documentation issue. And the reviewer there can take care of that

Tarun-kishore:
> I think we can specify which release in the documentation issue. And the reviewer there can take care of that

The change will be available in the next release after merging, correct?

Let me know if there are some changes needed in PR after fixing the CI builds.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/Transition.kt
```diff
@@ -110,6 +114,8 @@ data class Conditions(
         size = sin.readOptionalWriteable(::ByteSizeValue),
         cron = sin.readOptionalWriteable(::CronSchedule),
         rolloverAge = sin.readOptionalTimeValue(),
+        noAlias = sin.readOptionalBoolean(),
```
bowenlan-amzn, line 117: We would need to wrap this in version check for backward compatibility. sth like
https://github.com/opensearch-project/index-management/blob/d8f83ddc13bae7d61aae1e073ce73160cbcc8f6d/src/main/kotlin/org/opensearch/indexmanagement/rollup/model/ISMRollup.kt#L147-L150

Same for the write

bowenlan-amzn:
@vikasvb90 Please review this PR of enhancement in ISM transition.

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/TransitionActionIT.kt
```diff
@@ -147,4 +147,132 @@ class TransitionActionIT : IndexStateManagementRestTestCase() {
         // Should have evaluated to true
         waitFor { assertEquals(AttemptTransitionStep.getSuccessMessage(indexName, secondStateName), getExplainManagedIndexMetaData(indexName).info?.get("message")) }
     }
+
+    fun `test noAlias transition condition`() {
```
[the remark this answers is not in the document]
Tarun-kishore: Why would this change bwc though, It isn't modifying any existing conditions, just adding two new conditions.

src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/util/ManagedIndexUtilsTests.kt
```diff
@@ -243,7 +243,15 @@ class ManagedIndexUtilsTests : OpenSearchTestCase() {
         assertTrue(
             "No conditions should pass",
             emptyTransition
-                .evaluateConditions(indexCreationDate = Instant.now(), numDocs = null, indexSize = null, transitionStartTime = Instant.now(), rolloverDate = null),
```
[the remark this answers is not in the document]
Tarun-kishore: Added unit tests for both conditions, Integration tests are already implemented

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/model/Transition.kt
bowenlan-amzn: Next release would be 3.2.

The JSON parsing logic doesn't seem to be affected by newly added field. 
So for example, 3.1 node send 3.2 node of this transition data, it doesn't have noAlias write/serialized at all, but 3.2 node still try to readOptionalBoolean, which would fail I believe.

lanbowen23: They are same. Nodes of certain OS version can only use plugin built with that version.

vikasvb90:
Overall change looks good to me. Let me know if you plan to add a bwc test.

Tarun-kishore:
Adding bwc test makes sense for this change, I'll add it.
I'm getting access denied error when downloading tar for bwc test though, not sure if it's an issue with my network or some gradle issue.

Tarun-kishore:
@vikasvb90 I've added bwc test for ISM policy.

Tarun-kishore:
Hi @bowenlan-amzn @vikasvb90, just a quick nudge — I’ve addressed all the requested changes. Would appreciate a review when you get a chance. Thanks!

vikasvb90 (approved):
Thanks @Tarun-kishore for adding BWC test! Approving PR. @bowenlan-amzn Can you take another pass and approve?

PR #1442, opened 2025-07-07
Fix Integration test and lint errors

### LINKED ISSUE #1441: [BUG] RollupInterceptorIT integration test is failing
**What is the bug?**  
The `RollupInterceptorIT` integration test is failing due to a mismatch in the expected error message order when validating rollup search failure. The actual message contains the same fields, but the order of "missing field" entries is not deterministic. 

1. Run `RollupInterceptorIT.test roll up search` in the Index Management module.  
2. Test fails with different order of missing fields in the error string.

The error message should clearly report the required missing fields, but the test should not be sensitive to their order if all required fields are present in the message.

 - OS: [MacOS]
 - Version [failure reproducable in commit `ca0416c4e9986ea1b23d643031702e9e30e16a65`]
 - Plugins index-management

Image

Add any other context about the problem.

Tarun-kishore:
### Description

This PR fixes a failure in the `RollupInterceptorIT.test roll up search` integration test.

The test was asserting the exact error message returned when no suitable rollup job was found for a query. However, the list of missing fields in the error message is unordered, resulting in non-deterministic string output (`"missing field timestamp"` appearing before or after `"missing field RateCodeID"`).
image

This caused the test to fail intermittently due to strict string comparison.

This PR also fixes some `ktlint` errors by running `ktlint --format`

---

### Fix

- Updated the test assertion to validate the **presence** of expected substrings individually using `assertTrue()` instead of relying on exact ordering via `assertEquals()`.
- Ensures test stability while preserving correctness.

---

### Testing

- Ran `RollupInterceptorIT` locally to confirm consistent pass.
- Verified that error message contains all expected fields:
  - `missing field RateCodeID`
  - `missing field timestamp`
  - `missing sum aggregation on total_amount`
  
image

---

### Related Issues
Resolves #1441 

### Check List

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateRollover.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/validation/ValidateRollover.kt
@@ -31,8 +31,7 @@ class ValidateRollover(
         if (skipRollover(indexName) || alreadyRolledOver(rolloverTarget, indexName)) return this
 
         if (!isDataStream) {
-            if (!hasAlias(rolloverTarget, indexName) || !isWriteIndex(rolloverTarget, indexName)
-            ) {
+            if (!hasAlias(rolloverTarget, indexName) || !isWriteIndex(rolloverTarget, indexName)) {
                 return this
             }
         }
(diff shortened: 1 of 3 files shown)


cwperks:
Thank you @Tarun-kishore ! I hit the same failure in https://github.com/opensearch-project/index-management/pull/1438 and included a fix in there, but its a much larger PR.

cwperks (approved):

bowenlan-amzn (approved):

PR #1484, opened 2025-09-03
Run tests with custom user attribute serialization

cwperks:
### Description

This PR adds a new mapping to the ISM system indices for `user_attributes` that was added in https://github.com/opensearch-project/common-utils/pull/827

This PR runs tests with a user that has custom attributes and verifies that persisting a copy of the user works correctly in ISM system indices.

### Related Issues

Working towards a solution for https://github.com/opensearch-project/alerting/issues/1829

### Check List

### DIFF
--- a/src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
+++ b/src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
@@ -41,7 +41,7 @@ import javax.management.remote.JMXConnectorFactory
 import javax.management.remote.JMXServiceURL
 
 abstract class IndexManagementRestTestCase : ODFERestTestCase() {
-    val configSchemaVersion = 24
+    val configSchemaVersion = 25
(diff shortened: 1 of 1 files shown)


src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/IndexStateManagementRestTestCase.kt
```diff
@@ -88,6 +88,66 @@ abstract class IndexStateManagementRestTestCase : IndexManagementRestTestCase()
     val explainResponseOpendistroPolicyIdSetting = "index.opendistro.index_state_management.policy_id"
     val explainResponseOpenSearchPolicyIdSetting = "index.plugins.index_state_management.policy_id"
 
+    @Before
```
bowenlan-amzn, line 91: Could we just do this in one test? If the goal is to make sure an user's custom attribute can be persisted and updated in ISM system index? 
Current way seems will apply to all the tests.

PR #1507, opened 2025-10-13
Use optionalField for creation in ExplainSMPolicy serialization

### LINKED ISSUE #1506: [BUG] ExplainSMPolicy incorrectly uses `.field()` for nullable creation field
**What is the bug?**
In `ExplainSMPolicy.kt`, the `toXContent()` method incorrectly uses `.field()` for the nullable `creation` field instead of `.optionalField()`. This is inconsistent with how the `deletion` field is handled on the same line and with the reference implementation in `SMMetadata.toXContent()`. Using `.field()` with a nullable value can cause incorrect serialization behavior when the creation workflow metadata is null.

1. Create a snapshot management policy where the creation workflow metadata can be null
2. Call the Explain API to retrieve the policy metadata
3. Observe the JSON response when `metadata.creation` is null
4. See incorrect serialization behavior due to `.field()` being used instead of `.optionalField()`

The `creation` field should use `.optionalField()` to properly handle null values, just like the `deletion` field does. When `creation` is null, it should be omitted from the JSON response or handled gracefully rather than causing potential serialization issues.

Expected code (line 39 in ExplainSMPolicy.kt):
```kotlin
.optionalField(SMMetadata.CREATION_FIELD, it.creation)
```

Current (incorrect) code:
```kotlin
.field(SMMetadata.CREATION_FIELD, it.creation)
```

 - OS: N/A (Code-level bug in serialization)
 - Version: 3.3.0-SNAPSHOT
 - Plugins: OpenSearch Index Management Plugin - Snapshot Management component

N/A - This is a code-level serialization bug

- **Location**: `src/main/kotlin/org/opensearch/indexmanagement/snapshotmanagement/model/ExplainSMPolicy.kt`, line 39
- **Inconsistency**: The `deletion` field on line 40 correctly uses `.optionalField()`
- **Reference**: `SMMetadata.kt` line 55 shows the correct implementation using `.optionalField()` for the creation field
- **Field Type**: `creation` is declared as nullable `WorkflowMetadata?` in `SMMetadata` (line 42 of SMMetadata.kt)
- **Root Cause**: The creation field was recently changed to be optional (nullable) as part of version 3.3.0 changes, but the ExplainSMPolicy serialization was not updated accordingly

Tarun-kishore:
### Description
Fixed incorrect serialization of the `creation` field in `ExplainSMPolicy.toXContent()`. The method was using `.field()` instead of `.optionalField()` for the nullable `creation` field, which is inconsistent with how the `deletion` field is handled and with the reference implementation in `SMMetadata.toXContent()`.

**Changed line 39 in `ExplainSMPolicy.kt`:**
- **Before**: `.field(SMMetadata.CREATION_FIELD, it.creation)`
- **After**: `.optionalField(SMMetadata.CREATION_FIELD, it.creation)`

This ensures proper handling of null creation workflow metadata in Explain API responses and aligns with the serialization behavior in `SMMetadata`.

### Related Issues
Resolves #1506 

### Check List

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/ISMTemplateService.kt
@@ -22,22 +22,44 @@ import org.opensearch.indexmanagement.util.IndexManagementException
 @Suppress("ComplexMethod")
 fun validateFormat(indexPatterns: List<String>): OpenSearchException? {
     val indexPatternFormatErrors = mutableListOf<String>()
+    var hasInclusionPattern = false
+
     for (indexPattern in indexPatterns) {
-        if (indexPattern.contains("#")) {
+        // Strip the exclusion prefix (-) if present for validation
+        val isExclusionPattern = indexPattern.startsWith("-")
+        if (!isExclusionPattern) {
+            hasInclusionPattern = true
+        }
+        val patternToValidate = if (isExclusionPattern) {
+            indexPattern.substring(1)
+        } else {
+            indexPattern
+        }
+
+        // Check if exclusion pattern is empty after removing the prefix
+        if (isExclusionPattern && patternToValidate.isEmpty()) {
+            indexPatternFormatErrors.add("index_pattern [$indexPattern] must have content after '-' exclusion prefix")
+        }
+        if (patternToValidate.contains("#")) {
             indexPatternFormatErrors.add("index_pattern [$indexPattern] must not contain a '#'")
         }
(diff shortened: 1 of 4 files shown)


Tarun-kishore:
I've added unit test covering the parsing, but unit test cannot change `Version.Current` value. I can add a method like `getCurrentVersion` and mock that but that doesn't seem like a good way. I've also updated bwc tests to get explain response, Seems like it's coverage isn't included in coverage check.
I've update version check to `onOrAfter V.3_3_0` to test bwc coverage, once `V.3_4_0` is available I'll update it to that.

PR #1517, opened 2025-10-25
Adapt ActionFilter interface implementation to core change

nibix:
### Description

The core PR https://github.com/opensearch-project/OpenSearch/pull/18523 has changed the `ActionFilter` interface by adding a new `apply()` method with an additional parameter. The old `apply()` method is now deprecated. This change adds the parameter to the ActionFilter implementations.

### Check List


Hailong-am:
please check the complie error

nibix:
The error occurs because index-management still uses OpenSearch core in version `3.3.0-SNAPSHOT` according to `build.gradle`; this can be only applied when it is bumped to `3.4.0-SNAPSHOT`. I am moving this PR to draft status until this is changed.

nibix:
Note: Merging #1499 should unblock this

PR #1527, opened 2025-11-05
Add support for include_aliases, ignore_index_settings, and number_of_replicas in convert_index_to_remote action

### LINKED ISSUE #21093: [BUG] Convert index to remote not deleting old index
### Describe the bug

Hi i have this running policy:

```
{
  "policy": {
    "policy_id": "archive-to-remote",
    "description": "Archive indices older than 2 days to S3 searchable snapshots, then delete local index",
    "default_state": "hot",
    "states": [
      {
        "name": "hot",
        "actions": [],
        "transitions": [
          {
            "state_name": "archive",
            "conditions": {
              "min_index_age": "2d"
            }
          }
        ]
      },
      {
        "name": "archive",
        "actions": [
          {
            "retry": {
              "count": 3,
              "backoff": "exponential",
              "delay": "1m"
            },
            "snapshot": {
              "repository": "snapshot-repo",
              "snapshot": "{{ctx.index}}"
            }
          },
          {
            "retry": {
              "count": 3,
              "backoff": "exponential",
              "delay": "1m"
            },
            "convert_index_to_remote": {
              "repository": "snapshot-repo",
              "snapshot": "{{ctx.index}}"
            }
          }
        ],
        "transitions": [
          {
            "state_name": "delete",
            "conditions": {
              "min_index_age": "3d"
            }
          }
        ]
      },
      {
        "name": "delete",
        "actions": [
          {
            "retry": {
              "count": 3,
              "backoff": "exponential",
              "delay": "1m"
            },
            "delete": {}
          }
        ],
        "transitions": []
      }
    ],
    "ism_template": [
      { "index_patterns": ["logs-*"], "priority": 100 },
      { "index_patterns": ["events-*"], "priority": 100 },
      { "index_patterns": ["traces-*"], "priority": 100 },
      { "index_patterns": ["oci-logs-*"], "priority": 100 },
      { "index_patterns": ["security-auditlog-*"], "priority": 100 }
    ]
  }
}
```

The new indexes are well created but the old ones are not deleted.
I understood the convert_index_to_remote would delete the old index after creating the new one but thats not the case thats why i added the delete action.

Image

### Related component

Indexing

### Expected behavior

Old index deleted

### Additional Details

**Plugins**
Please list all plugins currently enabled.

**Screenshots**

**Host/Environment (please complete the following information):**

**Additional context**

wntmddus:
## Description

Extends `convert_index_to_remote` ISM action with additional restore options and safety controls.

### New fields

| Field | Type | Default | Description |
|-------|------|---------|-------------|
| `include_aliases` | boolean | `false` | Restore aliases from the snapshot |
| `ignore_index_settings` | string | `""` | Comma-separated settings to ignore during restore |
| `number_of_replicas` | integer | *not set* | Override replica count. Omit to inherit from snapshot |
| `delete_original_index` | boolean | `false` | Delete the local index after remote index is confirmed |
| `rename_pattern` | string | `$1_remote` | Name pattern for the remote index. Must contain `$1` |

### Behavior

1. Step initiates restore with `waitForCompletion=false`, returns `CONDITION_NOT_MET`
2. On re-execution, checks if remote index exists (verified via `index.store.type = remote_snapshot`)
3. If `delete_original_index = true`, deletes the original — step **fails** if deletion fails (retryable)
4. Marks `COMPLETED` once remote index is confirmed and original is cleaned up (if requested)

### Safety

- **Remote index validation**: Before trusting a pre-existing remote index, verifies it has `store.type = remote_snapshot` (prevents data loss from stale/unrelated indices with the same name)
- **Delete is idempotent**: If original index is already gone on retry, succeeds without error
- **Rename validation**: `rename_pattern` must be non-empty and contain `$1`; fails fast otherwise
- **Replica count**: Omitting `number_of_replicas` preserves the snapshot's original setting (no silent override)

### Version compatibility

New fields are gated behind `V_3_7_0`. Stream serialization uses `-1` as sentinel for "unset" replicas. `rename_pattern` remains gated at `V_3_5_0`.

Schema version bumped from 27 → 28.

## Related issue

Resolves https://github.com/opensearch-project/OpenSearch/issues/21093

## Test plan

- [x] Unit tests for XContent and stream round-trip (all fields, individually and combined)
- [x] Backward/forward compatibility tests (V_3_4_0, V_3_5_0, V_3_7_0)
- [x] Unit tests for `rename_pattern` validation (blank, missing `$1`)
- [x] Integration test: snapshot → convert → verify remote index exists
- [x] Integration test: `delete_original_index = true` → verify original is deleted
- [x] Integration test: custom `rename_pattern`


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/restore/AttemptRestoreStep.kt
```diff
@@ -127,6 +122,87 @@ class AttemptRestoreStep(private val action: ConvertIndexToRemoteAction) : Step(
...
+            .build()
+        restoreSnapshotRequest.indexSettings(indexSettings)
+
+        val response: RestoreSnapshotResponse = context.client.admin().cluster().suspendUntil {
+            restoreSnapshot(restoreSnapshotRequest, it)
+        }
+
+        when (response.status()) {
+            RestStatus.ACCEPTED, RestStatus.OK -> {
+                deleteOriginalIndex(context, indexName, mutableInfo)
```
bowenlan-amzn: Since the snapshot restore is not waiting for completion, I worry delete index here would disrupt the search. 

I see you mentioned this:
> the action automatically deletes the original index once the restore operation is accepted (status: ACCEPTED or OK), ensuring a clean conversion process where only the remote index remains.

But I think we should keep this an option for user, and disabled by default.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/ConvertIndexToRemoteAction.kt
```diff
@@ -25,8 +29,18 @@ class ConvertIndexToRemoteAction(
...
         const val SNAPSHOT_FIELD = "snapshot"
+        const val INCLUDE_ALIASES_FIELD = "include_aliases"
+        const val IGNORE_INDEX_SETTINGS_FIELD = "ignore_index_settings"
+        const val NUMBER_OF_REPLICAS_FIELD = "number_of_replicas"
+        const val DELETE_ORIGINAL_INDEX_FIELD = "delete_original_index"
         const val RENAME_PATTERN_FIELD = "rename_pattern"
         const val DEFAULT_RENAME_PATTERN = "\$1_remote"
+
+        // Version when new fields (includeAliases, ignoreIndexSettings, numberOfReplicas, deleteOriginalIndex) were added
+        val VERSION_WITH_NEW_FIELDS = Version.V_3_3_0
```
shiv0408: Why are we using this version?

My concern for adding the version was bwc during version upgrade. Suppose during upgrade, the request is processed on one of the nodes running on the upgraded version with this code and the Cluster Manager where the request needs to be completed is not upgraded yet, it will not be able to parse the input stream sent from the node with upgraded version. 

Please use the version in which this PR will be merged for these checks.

shiv0408 (changes requested):

spapadop:
> @wntmddus the version needs to be updated again to 3.6 now. Hopefully this will then make it in because the documentation is now wrong because it already includes those settings. ([opensearch-project/documentation-website#11495](https://github.com/opensearch-project/documentation-website/pull/11495)).

I guess that should rather be `3.7` now.

wntmddus:
> > @wntmddus the version needs to be updated again to 3.6 now. Hopefully this will then make it in because the documentation is now wrong because it already includes those settings. ([opensearch-project/documentation-website#11495](https://github.com/opensearch-project/documentation-website/pull/11495)).
> 
> I guess that should rather be `3.7` now.

build is failing with 3.7 so ill have to switch back to 3.6? @spapadop

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/restore/AttemptRestoreStep.kt
```diff
@@ -128,6 +125,92 @@ class AttemptRestoreStep(private val action: ConvertIndexToRemoteAction) : Step(
...
+
+        val response: RestoreSnapshotResponse = context.client.admin().cluster().suspendUntil {
+            restoreSnapshot(restoreSnapshotRequest, it)
+        }
+
+        when (response.status()) {
+            RestStatus.ACCEPTED, RestStatus.OK -> {
+                // Only delete original index if explicitly enabled
+                if (action.deleteOriginalIndex) {
+                    deleteOriginalIndex(context, indexName, mutableInfo)
```
shiv0408: This was flagged by @bowenlan-amzn earlier as well. As we are not waiting for completion, we might delete the original before the restore is completed.
Why do we want to support deleting the original index during the same step?

Also, as we are keeping wait for completion as false, this step will complete even if the index in not restored. And having another following step to delete the index might also land up in the same situation.
We can take another field input from user to wait for completion for this step and that should be true for delete original index to be true.

wntmddus: fixed it

sashalarsoo:
Hey, Are the number of replicas already working on opensearch because i have problems to make it work and i see this pr still open ? I also have problem after the remote index is created the original index is not deleted.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/restore/AttemptRestoreStep.kt
wntmddus: @shiv0408

shiv0408: @wntmddus Sorry, can you please point out what changes did you make to fix this?

bowenlan-amzn (dismissed):

bowenlan-amzn (approved):

PR #1565, opened 2026-01-04
Add `delete_snapshot` option to ISM delete action for cleaning up searchable snapshot sources

### LINKED ISSUE #1476: [FEATURE] Add "delete searchable_snapshot" in ISM policies
The goal is to also cleanup the data from *remote snapshot repository*, if the data has *aged out*.

Currently, we can "only" delete *indices* which might still be on *local* disks or *mounted* via a *searchable_snapshot* (currently, index names following `$1_remote` pattern).

But deleting a *mounted* index does *not* also delete the corresponding *snapshot* on the remote repository. This means, the remote repository will *never* be cleaned up, even if the ISM policy states, that the data is *too old* to be kept.

Or is this wrong?

We want to push *older data* early into the *warm* tier (using *searchable_snapshots*) and then, having those snapshots *queryable* for a longer amount of time.

Adding an additional option, like `delete_snapshot` to the `delete` action to support to also delete the *snapshot*, that was used to *mount* the remote index

```
actions": [
    {
        "retry": {
            "count": 3,
            "backoff": "exponential",
            "delay": "1m"
        },
        "delete": {
          "delete_snapshot": true|false
        }
```

Currently, none. Could only think of some kind of cronjob, checking the age of all `*_remote` indices and if the *index* is too old, to delete the index + delete the corresponding snapshot. But this would also mean, we would need to have some kind of *map* in case you want to use different kind of *retentions* for different indices (what we definitely do :-/ )

It is basically [this](https://www.elastic.co/docs/reference/elasticsearch/index-lifecycle-actions/ilm-delete)

lawofcycles:
### Description
This PR adds a new optional `delete_snapshot` parameter to the ISM delete action. When enabled, it automatically deletes the source snapshot of a searchable snapshot index along with the index itself.

**use case**
Hot → Warm (searchable snapshot) → Delete lifecycle where both the index and underlying snapshot should be removed to free up storage.

**format**
```json
"actions": [
    {
        "delete": {
            "delete_snapshot": true
        }
    }
],
```

**behavior**
- If `delete_snapshot: true` and the index is a searchable snapshot, both the index and its source snapshot are deleted
- If the snapshot is still used by other indices, snapshot deletion is skipped (index deletion still succeeds)
- If the index is not a searchable snapshot, only the index is deleted (no error)
- Default is `false` for backward compatibility

In addition to unit tests, I manually tested with local build.

### Related Issues
Resolves #1476

### Check List

### DIFF
--- a/.github/workflows/test-and-build-workflow.yml
+++ b/.github/workflows/test-and-build-workflow.yml
@@ -52,7 +52,7 @@ jobs:
       # This is a hack, but this step creates a link to the X: mounted drive, which makes the path
       # short enough to work on Windows
       - name: Build with Gradle
-        timeout-minutes: 20
+        timeout-minutes: 25
         run: |
           chown -R 1000:1000 `pwd`
           su `id -un 1000` -c "./gradlew build ${{ env.TEST_FILTER }}"
--- a/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/DeleteAction.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/DeleteAction.kt
@@ -5,6 +5,9 @@
 
 package org.opensearch.indexmanagement.indexstatemanagement.action
 
+import org.opensearch.core.common.io.stream.StreamOutput
+import org.opensearch.core.xcontent.ToXContent
+import org.opensearch.core.xcontent.XContentBuilder
 import org.opensearch.indexmanagement.indexstatemanagement.step.delete.AttemptDeleteStep
 import org.opensearch.indexmanagement.spi.indexstatemanagement.Action
 import org.opensearch.indexmanagement.spi.indexstatemanagement.Step
@@ -12,15 +15,30 @@ import org.opensearch.indexmanagement.spi.indexstatemanagement.model.StepContext
 
 class DeleteAction(
     index: Int,
+    val deleteSnapshot: Boolean = false,
 ) : Action(name, index) {
     companion object {
         const val name = "delete"
+        const val DELETE_SNAPSHOT_FIELD = "delete_snapshot"
     }
 
-    private val attemptDeleteStep = AttemptDeleteStep()
+    private val attemptDeleteStep = AttemptDeleteStep(this)
     private val steps = listOf(attemptDeleteStep)
 
     override fun getStepToExecute(context: StepContext): Step = attemptDeleteStep
 
     override fun getSteps(): List<Step> = steps
+
+    override fun populateAction(builder: XContentBuilder, params: ToXContent.Params) {
+        builder.startObject(type)
+        if (deleteSnapshot) {
+            builder.field(DELETE_SNAPSHOT_FIELD, deleteSnapshot)
+        }
+        builder.endObject()
+    }
+
+    override fun populateAction(out: StreamOutput) {
+        out.writeInt(actionIndex)
+        out.writeBoolean(deleteSnapshot)
+    }
 }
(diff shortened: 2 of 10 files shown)


lawofcycles:
> Hi, please feel free to update the `timeout-minutes` in the workflow file to get the checks running to the end https://github.com/opensearch-project/index-management/blob/main/.github/workflows/test-and-build-workflow.yml

@bowenlan-amzn 
Thanks! CI passed after updating timeout-minutes to 25 minutes.

lawofcycles:
@bowenlan-amzn Thank you for the suggestion. Let me describe the use case I'm trying to address, so we can discuss whether the existing SM deletion-only policy can cover it.

**Use case (from #1476)**

A typical Hot -> Warm (searchable snapshot) -> Delete lifecycle managed by a ISM policy

1. ISM creates a snapshot of the index
2. ISM converts the index to a searchable snapshot (remote_snapshot) via `convert_index_to_remote`
3. After a retention period, ISM deletes the searchable snapshot index

The problem is that step 3 only deletes the index, not the underlying snapshot in the repository.

I looked into using an SM deletion-only policy with `snapshot_pattern` and `max_age` as an alternative. However, the current SM deletion logic (`DeletingState`) does not check whether a snapshot is still being used as the source of a searchable snapshot index. It filters snapshots solely by `max_age`/`max_count`/`min_count`.

This means that if the SM deletion schedule runs before ISM deletes the searchable snapshot index, SM could delete a snapshot that is still actively backing a mounted index, making that index's data inaccessible.

Since ISM controls the index lifecycle, it knows exactly when the index is no longer needed, so triggering the snapshot cleanup at the point of index deletion would avoid this timing issue.

I would appreciate your thoughts on how to best address this use case.

lawofcycles:
> 
> Your use case involves a 1:1 mapping between snapshots and indices. In ISM, we can implement this by storing the snapshot name in the managed index metadata after successful completion, then reading and deleting that specific snapshot during the delete step.
> 
> The snapshot management feature differs—it creates snapshots for index groups on a time-based schedule. Deletion retains only the latest snapshots (at minimum, the most recent one) rather than removing all snapshots. However, it's unclear whether this approach works for searchable indices—specifically, whether a searchable index must bind to one specific snapshot or can function with any snapshot containing it. Could you check this once you got time, thanks!

@bowenlan-amzn I looked into the binding behavior of searchable snapshot indices. Here are my findings from testing on OpenSearch 3.5.0.

**1. Searchable snapshots are hard bound to a specific snapshot**
When an index is restored with `storage_type: remote_snapshot`, the index settings record the exact snapshot it was created from:
```
index.searchable_snapshot.repository: my-repo
index.searchable_snapshot.snapshot_id.name: snapshot-a
index.searchable_snapshot.snapshot_id.uuid: c43AnLliStevfiMUBtzxkA
index.store.type: remote_snapshot
```

The binding is at the UUID level, so the index cannot function with a different snapshot, even if that snapshot contains the same data.

**2. OpenSearch protects the bound snapshot from deletion**

Attempting to delete `snapshot-a` while the searchable snapshot index exists returns.
```
snapshot_in_use_deletion_exception:
[my-repo:[snapshot-a]] These remote snapshots are backing some indices
and hence can't be deleted! No snapshots were deleted.
```

**3. SM deletion policy behavior with searchable snapshot bound snapshots**
I also tested an SM deletion only policy with `snapshot_pattern` configured to match the target snapshots. When SM attempts to delete a snapshot that is still backing a searchable snapshot index, it receives the same `snapshot_in_use_deletion_exception`. SM treats this as a retryable error, retries 3 times, then marks the workflow as `FAILED` and resets. On the next scheduled trigger, it repeats the same cycle. This continues until the searchable snapshot index is deleted by ISM, after which SM can successfully delete the snapshot on its next run.

givilleneuve:
Hey team,

Any updates on this one? It is been a while it has been opened and it can be a very major cost issue when using cloud storage for the searchable snapshots if they are not being purged.

rpasche:
Any updates?

PR #1568, opened 2026-01-12
Add optional `rename_pattern` parameter to convert_index_to_remote action

### LINKED ISSUE #1426: [BUG] The "convert_index_to_remote" operation restored  the target  index  suffixed  whith "_remote". In most cases, this new index will match the same policy as the original index, which can trigger the index management process again
**What is the bug?**
The "convert_index_to_remote" operation restored  the target  index  suffixed  whith "_remote". In most cases, this new index will match the same policy as the original index, which can trigger the index management process again (although it will  fail due to  read-only ). This is not what we want. To avoid this issue, we can use the "remote_" prefix instead of "_remote" suffix .

1.
```
{
  "policy": {
    "description": "remote",
    "default_state": "hot",
    "schema_version": 1,
    "states": [
      {
        "name": "hot",
        "actions": [
          {
            "rollover": {
              "min_index_age": "7d",
              "min_primary_shard_size": "30gb"
            }
          }
        ],
        "transitions": [
          {
            "state_name": "warm"
          }
        ]
      },
      {
        "name": "warm",
        "actions": [
          {
            "replica_count": {
              "number_of_replicas": 1
            }
          },
          {
            "allocation": {
              "require": {
                "temp": "warm"
              }
            }
          }
        ],
        "transitions": [
          {
            "state_name": "remote"
          }
        ]
      },
      {
        "name": "remote",
        "actions": [
          {
            "snapshot": {
                "repository": "my_backup",
                "snapshot": ""
            }, 
            "convert_index_to_remote": {
                "repository": "my_backup",
                "snapshot": ""
            }
          }
        ]
      }
    ],
    "ism_template": {
      "index_patterns": ["log*"],
      "priority": 100
    }
  }
}
```
2. origin index: log_0001
3. remote index: log_0001_remote  which will trigger this policy too!

if we use the name “remote_log_0001” which will  not trigger the same policy as the original index

 - OS: ubuntu22
 - Version 3.0.0

Add any other context about the problem.

lawofcycles:
### Description

This PR adds a new optional rename_pattern parameter to the ISM convert_index_to_remote action. This allows users to customize the naming pattern for the restored searchable snapshot index.

### Use Case

Users want to control the naming convention when converting indices to searchable snapshots, such as adding a prefix (remote_) instead of the default suffix (_remote).

### Format

```json
"actions": [
    {
        "convert_index_to_remote": {
            "repository": "my_backup",
            "snapshot": "{{ctx.index}}",
            "rename_pattern": "remote_$1"
        }
    }
]
```

### Behavior

- $1 is replaced with the original index name
- Default is $1_remote for backward compatibility
- Example: index my-index with rename_pattern: "remote_$1" becomes remote_my-index

### Testing

- Unit tests for default/custom values, serialization, and parsing
- Integration test verifying custom rename pattern creates correctly named index
- Manually tested with local cluster

### Related Issues
Resolves #1426

### Check List

- [x] New functionality has been documented. https://github.com/opensearch-project/documentation-website/pull/11866
- [ ] API changes companion pull request created.
- [x] Commits are signed per the DCO using --signoff.
- [x] Public documentation issue/PR created.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/rollup/RollupIndexer.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/rollup/RollupIndexer.kt
@@ -14,6 +14,7 @@ import org.opensearch.action.bulk.BulkRequest
 import org.opensearch.action.bulk.BulkResponse
 import org.opensearch.action.index.IndexRequest
 import org.opensearch.cluster.service.ClusterService
+import org.opensearch.common.io.stream.BytesStreamOutput
 import org.opensearch.common.settings.Settings
 import org.opensearch.common.xcontent.XContentType
 import org.opensearch.core.rest.RestStatus
@@ -29,6 +30,7 @@ import org.opensearch.indexmanagement.util.IndexUtils.Companion.ODFE_MAGIC_NULL
 import org.opensearch.indexmanagement.util.IndexUtils.Companion.hashToFixedSize
 import org.opensearch.search.aggregations.bucket.composite.InternalComposite
 import org.opensearch.search.aggregations.metrics.InternalAvg
+import org.opensearch.search.aggregations.metrics.InternalCardinality
 import org.opensearch.search.aggregations.metrics.InternalMax
 import org.opensearch.search.aggregations.metrics.InternalMin
 import org.opensearch.search.aggregations.metrics.InternalSum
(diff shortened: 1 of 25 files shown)


src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/ConvertIndexToRemoteActionParser.kt
```diff
@@ -18,13 +20,15 @@ class ConvertIndexToRemoteActionParser : ActionParser() {
     override fun fromStreamInput(sin: StreamInput): Action {
         val repository = sin.readString()
         val snapshot = sin.readString()
+        val renamePattern = sin.readString()
```
bowenlan-amzn: Same as write stream output, need a version check

PR #1635, opened 2026-05-06
Fix _refresh_search_analyzers blocked on indices with METADATA_WRITE block

gaurav2612gupta:
### Description

`_refresh_search_analyzers` fails with HTTP 403 (`ClusterBlockException`) on any index that has a `METADATA_WRITE`-level block, such as CCR follower indices.

The root cause is that `TransportRefreshSearchAnalyzerAction` declares it requires `METADATA_WRITE` permission via `checkGlobalBlock`/`checkRequestBlock`, but the action only reloads synonym files from disk into shard-local memory — it does not mutate index metadata or cluster state.

### Changes

- Remove `ClusterBlockLevel.METADATA_WRITE` checks from `checkGlobalBlock` and `checkRequestBlock` (return `null`)
- Remove unused `ClusterBlockLevel` import
- Add integration test verifying refresh succeeds on an index with a read-only block

### Testing

- Added `test refresh succeeds with metadata write block` in `RefreshSearchAnalyzerActionIT`
- Existing tests pass (synonym reload on normal indices unaffected)

### Check List

- [x] New functionality includes testing
- [x] New functionality has been documented above
- [x] Commits are signed off (DCO)

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/refreshanalyzer/TransportRefreshSearchAnalyzerAction.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/refreshanalyzer/TransportRefreshSearchAnalyzerAction.kt
@@ -10,7 +10,6 @@ import org.opensearch.action.support.ActionFilters
 import org.opensearch.action.support.broadcast.node.TransportBroadcastByNodeAction
 import org.opensearch.cluster.ClusterState
 import org.opensearch.cluster.block.ClusterBlockException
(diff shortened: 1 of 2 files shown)


gaurav2612gupta (commented):
Changes -  
- Remove ClusterBlockLevel.METADATA_WRITE checks from checkGlobalBlock and checkRequestBlock (return null)
- Remove unused ClusterBlockLevel import
- Add integration test verifying refresh succeeds on an index with a read-only block

soosinha (approved):

PR #1667, opened 2026-06-12
Support mixed AND/OR rollover conditions via any_of

### LINKED ISSUE #1540: [FEATURE] Support AND operations on Rollover conditions

Currently, rollover conditions in ISM only support an OR relationship. This means a rollover is triggered as soon as any condition (age, size, docs, etc.) is met.

Users who want rollover to occur only when multiple conditions are satisfied together (e.g., both index age AND index size) cannot express this directly.

To work around this limitation, users add additional states and transition conditions in their ISM policy to enforce an AND-like behavior. This leads to more complex policies, reduced readability, and increased maintenance overhead.

Add native support for AND operations between rollover conditions, enabling users to specify that rollover should occur only when all specified conditions are met. We can introduce an additional flag possibly. which could help identify if the customer wants the conditions to be evaluated using AND or OR logic

**State-based workaround:**
Adding a separate state with transition conditions to emulate an AND behavior. This works but is cumbersome and unintuitive.

**External automation:**
Using scripts or monitoring systems to check multiple conditions and trigger manual rollover. This adds operational overhead and removes the convenience of built-in ISM logic.

- Customers often want to ensure both an index age threshold and a size threshold are met before rollover, especially in ingestion-heavy or cost-sensitive environments.
- Supporting AND logic makes ISM policies cleaner and more aligned with typical lifecycle management expectations found in similar systems.
- The proposed flag makes the feature opt-in and backward compatible.

praladhe:
### Description
Adds support for **mixed AND/OR rollover conditions** in the ISM `rollover` action through a
new, opt-in grouped syntax (`any_of`).

Today the rollover action's four conditions (`min_index_age`, `min_size`, `min_doc_count`,
`min_primary_shard_size`) are combined with pure OR — rollover triggers as soon as any single
condition is met. There is no way to express "roll over when (age >= 7d AND size >= 50gb) OR
(doc_count >= 100000000)".

This change introduces an alternative **OR-of-ANDs** form. The action can hold an ordered list
of condition groups under `any_of`; the conditions inside a group are AND-ed and the groups are
OR-ed. Rollover triggers when any one group is fully satisfied. Because any boolean combination of
the four conditions can be rewritten as OR-of-ANDs, this covers all practical rollover policies.

Example:

```json
{
  "rollover": {
    "any_of": [
      { "min_index_age": "7d", "min_size": "50gb" },
      { "min_doc_count": 100000000 }
    ]
  }
}
````

**Behavior and compatibility**

- The existing flat form is unchanged and remains the default (implicit OR), so all existing policies behave exactly as before.
- The flat form and the grouped `any_of` form are mutually exclusive within a single rollover action; specifying both is rejected at parse/construction time.
- An empty `any_of`, an empty group, non-positive thresholds, and unknown fields inside a group are all rejected with informative errors.
- Stream serialization of the grouped data is version-gated (introduced in `3.7.0`), following the existing `prevent_empty_rollover` / `3.4.0` pattern, so mixed-version clusters operate safely — older nodes never receive groups and fall back to flat behavior.
- The `prevent_empty_rollover` empty-index guard is preserved unchanged and still runs before condition evaluation.
- The ISM explain output reports grouped criteria under an `any_of` array; flat reporting is unchanged.

**Changes**

- New `RolloverConditionGroup` model (parsing, validation, XContent + stream serialization).
- `RolloverAction`: new optional `conditionGroups` field, mutual-exclusivity/empty validation, grouped XContent output, and version-gated stream write; new `any_of` constant and `TARGET_VERSION`.
- `RolloverActionParser`: parse `any_of` from XContent and version-gated read from stream.
- `evaluateConditions` in `ManagedIndexUtils`: refactored to a shared per-condition comparison helper reused by both flat (OR) and grouped (OR-of-ANDs) paths, keeping flat behavior byte-identical.
- `AttemptRolloverStep`: report flat conditions or grouped `any_of` criteria.
- Config index mapping: added `any_of` and bumped `schema_version` 29 → 30 (mapping and cached test copy kept in sync; test `configSchemaVersion` updated).

**Testing**

- Unit tests: 796 passing, including example tests and property-based tests (kotest) covering evaluation, serialization round-trips, version gating, and validation.
- Integration tests: `RolloverActionIT` (incl. a new end-to-end grouped `any_of` test) and `IndexManagementIndicesIT` pass against a real cluster.
- Verified manually on a live `3.7.0` cluster: a grouped policy rolled over when one group's conditions were satisfied while another group's were not.

**Public Documentation PR**: [#12740](https://github.com/opensearch-project/documentation-website/pull/12740)
**API changes companion pull request**: [#1158](https://github.com/opensearch-project/opensearch-api-specification/pull/1158)

### Related Issues
Resolves [#1540](https://github.com/opensearch-project/index-management/issues/1540)

### Check List
- [x] New functionality has been documented.
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

### CODE REVIEW

File: src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/RolloverAction.kt [hunk 1]
```diff
@@ -74,5 +94,8 @@ class RolloverAction(
         const val MIN_PRIMARY_SHARD_SIZE_FIELD = "min_primary_shard_size"
         const val COPY_ALIAS_FIELD = "copy_alias"
         const val PREVENT_EMPTY_ROLLOVER_FIELD = "prevent_empty_rollover"
+        const val ANY_OF_FIELD = "any_of"
+
+        val TARGET_VERSION: Version = Version.V_3_7_0
```
vishdivs: nit: lets remove TARGET_VERSION from the object used for defining field, Version.V3_7_0 is already a constant, lets use this directly

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/RolloverAction.kt
```diff
@@ -63,6 +79,10 @@ class RolloverAction(
         if (out.version.onOrAfter(Version.V_3_4_0)) {
             out.writeBoolean(preventEmptyRollover)
         }
+        if (out.version.onOrAfter(TARGET_VERSION)) {
+            out.writeBoolean(conditionGroups != null)
+            if (conditionGroups != null) out.writeList(conditionGroups)
+        }
```
vishdivs: what happens if a grouped policy is sent to a pre-3.7.0 node? suppose we created this policy in new version and Index was created before 3.7.0, what will the behaviour then?

praladhe: It never runs there. During a mixed-version upgrade, SkipExecution pauses all ISM (node versions differ), and the transport gate removes the groups before they reach the old node; if the old node parses the JSON directly it just rejects it (Invalid field: [any_of]). 
For index created before 3.7.0 it doesn't matter as ISM evaluates live index stats (age/size/docs), not the creation version. Once the cluster is on 3.7+, the grouped policy works exactly as normal.

vishdivs (approved):

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/RolloverAction.kt
```diff
@@ -63,6 +79,10 @@ class RolloverAction(
         if (out.version.onOrAfter(Version.V_3_4_0)) {
             out.writeBoolean(preventEmptyRollover)
         }
+        if (out.version.onOrAfter(Version.V_3_7_0)) {
```
tandonks, line 82: Since 3.7 is already released, we should change it to 3.8

praladhe: Agreed — if this ships in 3.8, V_3_8_0 is the right gate. I've gated on V_3_7_0 for now and will flip these to V_3_8_0 as soon as we bump the dependency to 3.8.0-SNAPSHOT.

tandonks (approved):
Cool looks good. Please make sure to followup on the documentation as this is a big change. Thanks for your first contribution :)

PR #1691, opened 2026-07-07
Implementation of allow actions in red cluster feature

skumarp7:
Add opt-in setting to run ISM jobs on a red cluster

Problem
-------
ISM does not execute managed-index jobs while the cluster health is red. When a
cluster turns red (for example due to disk pressure), delete policies that would
free up space never run, so the cluster cannot recover on its own without manual
intervention (issue #1127).

Approach
--------
Introduce an opt-in, dynamic cluster setting that lets operators allow ISM to keep
running while the cluster is red, while still protecting the cluster from actions
that would make a red cluster worse.

- New dynamic setting plugins.index_state_management.allow_running_on_red_cluster
  (NodeScope, Dynamic, default false) so existing behavior is unchanged and it can
  be toggled at runtime. It is registered in IndexManagementPlugin and tracked live
  in ManagedIndexRunner via addSettingsUpdateConsumer.

- Gate the existing red-cluster short-circuit in ManagedIndexRunner so jobs are
  skipped on a red cluster only when the setting is false (today's behavior). When
  the setting is true, execution proceeds.

- Safety guard: even when the override is enabled, a curated set of actions that
  create indices, require shard allocation, or add heavy I/O (force_merge,
  replica_count, shrink, snapshot, rollover, open, allocation,
  convert_index_to_remote, transform, rollup) are still skipped on a red cluster,
  because running them could further degrade an already red cluster. Recovery
  oriented actions such as delete continue to run so policies can help the cluster
  recover. This is expressed via RED_CLUSTER_RESTRICTED_ACTIONS and
  isActionAllowedOnRedCluster() in ManagedIndexSettings.

Tests
-----
- ManagedIndexSettingsTests: default value, dynamic node-scope properties, and the
  allowed vs. restricted action classification (delete allowed, resource-intensive
  actions restricted).
- IndexManagementSettingsTests: the new setting is returned by the plugin.
- ManagedIndexRunnerTests: registers the new setting so consumer wiring resolves.
- ManagedIndexRunnerIT: the setting round-trips through the cluster settings API.

Relates to #1127


skumarp7:
Hi @bowenlan-amzn, 

Please provide feedback :)

skumarp7:
Hi @vikasvb90 , @bowenlan-amzn , @Tarun-kishore 

Any feedback on the PR is welcomed :)

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/settings/ManagedIndexSettings.kt
```diff
@@ -18,11 +28,36 @@ class ManagedIndexSettings {
...
         const val DEFAULT_RESTRICTED_PATTERN = "\\.opendistro_security|\\.kibana.*|\\$INDEX_MANAGEMENT_INDEX"
         val ALLOW_LIST_NONE = emptyList<String>()
         val ALLOW_LIST_ALL = ISMActionsParser.instance.parsers.map { it.getActionType() }.toList()
         val SNAPSHOT_DENY_LIST_NONE = emptyList<String>()
 
+        // Actions that must never be executed while the cluster health is red, even when
+        // ALLOW_RUNNING_ON_RED_CLUSTER is enabled. These either create new indices, require
+        // shard allocation, or add significant I/O load, all of which can further degrade an
+        // already red cluster instead of helping it recover.
+        val RED_CLUSTER_RESTRICTED_ACTIONS: Set<String> =
```
shiv0408: I would rather have a list of allowed action and block all rest. Otherwise, any new action that is added would be allowed to run on red cluster by default.

skumarp7: Done.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/settings/ManagedIndexSettings.kt
```diff
@@ -18,11 +28,36 @@ class ManagedIndexSettings {
...
         val SNAPSHOT_DENY_LIST_NONE = emptyList<String>()
 
+        // List of actions that are allowed to run on a red cluster. Stored lowercased so that
+        // action name lookups can be done case-insensitively.
+        val RED_CLUSTER_ALLOWED_ACTIONS: Set<String> =
+            setOf(
+                DeleteAction.name,
+                CloseAction.name,
+                ReadOnlyAction.name,
+                ReadWriteAction.name,
```
Tarun-kishore: moving index to `read_write` state can potentially degrade the cluster stability, as indexing load will be increased on an existing red cluster. Should we consider skipping it?

skumarp7: Yes, it makes sense to remove read_write from the allow-list. Although the action itself is lightweight, it re-enables indexing, which can increase disk, CPU, memory, and recovery pressure on an already red cluster. Since it is not recovery-oriented, skipping it is the safer behavior.

skumarp7: @Tarun-kishore , have pushed a new commit with the changes. Please share your comments.

Tarun-kishore:
New changes lgtm.

Tarun-kishore:
@skumarp7 Can you fix the CI failures.

skumarp7:
@Tarun-kishore 

> @skumarp7 Can you fix the CI failures.

The CI seems to be failing due to a Kotlin compilation error in TransformRestTestCase.kt. The issue is with the deprecated RestClient class from the Java/OpenSearch client library. Should i supress the deprecation warnings ?

Tarun-kishore:
I checked the failure as well, those are because of RestClient deprecation warning. Since those are unrelated to this change, created a separate PR for that: https://github.com/opensearch-project/index-management/pull/1702.

Tarun-kishore:
@skumarp7 Can you add ITs for these changes.

skumarp7:
@Tarun-kishore, 


> @skumarp7 Can you add ITs for these changes.

Have added some ITs, let me know if they are okay?

Tarun-kishore (approved):

Tarun-kishore:
LGTM. Merging PR now.

PR #1695, opened 2026-07-09
Add original index name as alias in convert_index_to_remote action

### LINKED ISSUE #1694: [FEATURE] Add a new boolean parameter "add_original_name_as_alias" to convert_index_to_remote action

When using the `convert_index_to_remote` ISM action to convert indices to searchable snapshots, the restored remote index gets a different name (e.g., `remote_foo-bar-000001` via `rename_pattern`). This breaks existing index patterns and dashboards that relied on the original index name (e.g., `foo-bar-*` no longer resolves the converted index). Users are forced to either create additional index patterns to cover the new naming scheme or manually add aliases after conversion — neither of which scales well in automated lifecycle management.
Additionally, using `rename_pattern: "$1"` (same name) is not possible because the original index must still exist when the restore is initiated, causing a name conflict. And using `include_aliases: true` only restores aliases that existed at snapshot time — it does not create a new alias matching the original index name itself.

Add a new boolean parameter `add_original_name_as_alias` (default: `false`) to the `convert_index_to_remote` action. When set to `true` (requires `delete_original_index: true`), it automatically adds the original index name as an alias on the remote index after the original is deleted.

Example policy usage:
```
{
  "convert_index_to_remote": {
    "repository": "my-repo",
    "snapshot": "{{ctx.index}}",
    "include_aliases": true,
    "delete_original_index": true,
    "add_original_name_as_alias": true,
    "rename_pattern": "remote_$1"
  }
}
```

This ensures that `foo-bar-000001` becomes discoverable as an alias on `remote_foo-bar-000001`, so existing index patterns (`foo-bar-*`) continue to work transparently.

1. Using `rename_pattern: "$1"` (same name as original): Not feasible because the restore happens while the original index still exists, causing a name conflict.
2. Using `include_aliases: true` alone: Only restores aliases that existed at snapshot time. Does not create an alias matching the original index name itself — which is what index patterns typically resolve against.
3. External automation (scripts/cron): Running a separate process to add aliases after ISM completes the conversion. This works but adds operational complexity, is not atomic, and requires monitoring a separate system alongside ISM.
4. Creating additional index patterns: Adding patterns like remote_* to dashboards. This is a manual step, doesn't scale across teams, and defeats the purpose of transparent lifecycle management.

A related concern is the ISM re-application loop: if the original index matched a template like `foo-bar-*` that assigns an ISM policy, a restored index named `foo-bar-000001_remote` (default `rename_pattern`) would still match and re-enter the snapshot/convert cycle. Using `rename_pattern: "remote_$1"` avoids this (since `remote_foo-bar-000001` won't match `foo-bar-*`), but then requires this feature to maintain discoverability under the original name. A higher-priority index template matching `remote_*` with no ISM policy assignment completes the solution.

robertpaschedag:
### Description
Add a new boolean parameter `add_original_name_as_alias` to the `convert_index_to_remote` ISM action. When enabled (requires `delete_original_index: true`), it adds the original index name as an alias on the newly created remote (searchable snapshot) index after the original is deleted.
This enables transparent query continuity: indices converted to searchable snapshots remain discoverable under their original index name via alias, without requiring changes to index patterns or dashboards. This is particularly useful when the index still needs to be resolvable via its original name through existing index patterns (e.g., `foo-bar-*` matching alias `foo-bar-000001` on `remote_foo-bar-000001`).

Also, ISM policies being applied to e.g. `foo-bar-*` would also match the default remote index (with pattern `$1_remote`). So those *remote* indices would also start a new lifecycle.

So

```
{
  "convert_index_to_remote": {
    "repository": "my-repo",
    "snapshot": "{{ctx.index}}",
    "include_aliases": true,
    "delete_original_index": true,
    "add_original_name_as_alias": true,
    "rename_pattern": "remote-$1"
  }
}
```

should make the original index name available as *alias*, while the *remote* indices do not fall under the ISM policy (of the original index)

Also bumps `schema_version` to 31 to ensure existing clusters automatically receive the new field mapping on plugin startup.

### Related Issues
Resolves #1694

### Check List


robertpaschedag:
Tried to fix the high impact review.

robertpaschedag:
I think the failed check was caused by some timeout issue, or?

src/main/resources/mappings/opendistro-ism-config.json
shiv0408: bump `schema_version` to 31

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/action/ConvertIndexToRemoteAction.kt
```diff
@@ -33,10 +40,15 @@ class ConvertIndexToRemoteAction(
...
         const val NUMBER_OF_REPLICAS_FIELD = "number_of_replicas"
         const val DELETE_ORIGINAL_INDEX_FIELD = "delete_original_index"
+        const val ADD_ORIGINAL_NAME_AS_ALIAS_FIELD = "add_original_name_as_alias"
         const val RENAME_PATTERN_FIELD = "rename_pattern"
         const val DEFAULT_RENAME_PATTERN = "\$1_remote"
 
         val VERSION_WITH_RESTORE_OPTIONS = Version.V_3_7_0
+
+        // TODO: Replace Version.CURRENT with the actual version constant (e.g., Version.V_3_8_0) once it is defined in OpenSearch core
+        val VERSION_WITH_ADD_ORIGINAL_ALIAS = Version.CURRENT
```
shiv0408: We can target 3.8 version for this PR. Replace this with `Version.V_3_8_0`.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/restore/AttemptRestoreStep.kt
```diff
@@ -116,6 +118,15 @@ class AttemptRestoreStep(private val action: ConvertIndexToRemoteAction) : Step(
                         info = mutableInfo.toMap()
                         return this
                     }
+                    if (action.addOriginalNameAsAlias) {
```
shiv0408, line 114: Please change the delete - alias sequence to a single atomic swap.

As written, `deleteOriginalIndex()` commits first and `addOriginalNameAsAlias()` is a separate follow-up call. If delete succeeds and alias-add fails even after retries for some reason, we will land in a FAILED state and would need to manually resolve that. I recommend using the primitive `MetadataIndexAliasesService.applyAliasActions` method to remove the index first and add an alias in a single atomic step. This way name will never be orphaned.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/restore/AttemptRestoreStep.kt
```diff
@@ -238,19 +240,59 @@ class AttemptRestoreStep(private val action: ConvertIndexToRemoteAction) : Step(
...
+        if (!context.clusterService.state().metadata().hasIndex(originalIndexName)) {
+            logger.info("Original index [$originalIndexName] already deleted")
+            val metadata = context.clusterService.state().metadata()
+            val indexMetadata = metadata.index(remoteIndexName)
+            if (indexMetadata != null && indexMetadata.aliases.containsKey(originalIndexName)) {
+                logger.info("Alias [$originalIndexName] already exists on remote index [$remoteIndexName], nothing to do")
+                return true
+            }
+            // Original gone but alias missing - just add alias
+            return addAliasOnly(context, remoteIndexName, originalIndexName, mutableInfo)
```
shiv0408, line 259: With following action, we won't land in a state where index is deleted by alias is missing. But it's fine to keep it like this.

src/main/kotlin/org/opensearch/indexmanagement/indexstatemanagement/step/restore/AttemptRestoreStep.kt
robertpaschedag: addressed this in [c0ebc56](https://github.com/opensearch-project/index-management/pull/1695/commits/c0ebc5665974b0de54e244a6a717c9bbbeb24330)

shiv0408:
#1677 Needs to be merged for build to pass with `Version_3_8_0`

robertpaschedag:
branch updated. Gentle ping...

PR #1708, opened 2026-07-31
Ignore Security plugin audit log tasks in waitForRunningTasks

Hailong-am:
### Description
The Security plugin is installed even in without-security CI runs; its audit logger writes asynchronously to `security-auditlog-*` indices on every REST call. These writes can still be in flight when `wipeAllIndices` calls `waitForRunningTasks`, causing a spurious "There are still tasks running" failure unrelated to the test itself.

Fix: extend `isIgnorableTask` to also check the task description and skip tasks targeting `security-auditlog-*` indices.

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/index-management/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1709, opened 2026-07-31
[Backport 3.8] Ignore Security plugin audit log tasks in waitForRunningTasks

Hailong-am:
### Description
Backport of #1708 to `3.8`.

The Security plugin is installed even in without-security CI runs; its audit logger writes asynchronously to `security-auditlog-*` indices on every REST call. These writes can still be in flight when `wipeAllIndices` calls `waitForRunningTasks`, causing a spurious "There are still tasks running" failure unrelated to the test itself.

Fix: extend `isIgnorableTask` to also check the task description and skip tasks targeting `security-auditlog-*` indices.

### Check List

### DIFF
--- a/src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
+++ b/src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
@@ -364,7 +364,8 @@ abstract class IndexManagementRestTestCase : ODFERestTestCase() {
                 for ((_, value1) in nodeTasks!!) {
                     val task = value1 as Map<String, Any>
                     val action = task["action"] as String
-                    if (isIgnorableTask(action)) continue
+                    val description = task["description"] as? String ?: ""
+                    if (isIgnorableTask(action, description)) continue
                     // runningTasks.add(task["action"].toString() + " | " + task["description"].toString())
                     runningTasks.add(task.toString())
                 }
@@ -372,12 +373,15 @@ abstract class IndexManagementRestTestCase : ODFERestTestCase() {
             return runningTasks
         }
 
-        // Ignore the task list API and segment replication background tasks (always running on Remote Store clusters)
-        private fun isIgnorableTask(action: String): Boolean =
+        // Ignore the task list API, segment replication background tasks, and Security plugin audit log
+        // writes. The Security plugin is installed even in without-security runs; its audit logger fires
+        // asynchronously on REST calls and can outlive the test that triggered them.
+        private fun isIgnorableTask(action: String, description: String = ""): Boolean =
             action == ListTasksAction.NAME ||
                 action == ListTasksAction.NAME + "[n]" ||
                 action.startsWith("segrep_") ||
-                action == "indices:admin/publishCheckpoint[p]"
+                action == "indices:admin/publishCheckpoint[p]" ||
+                description.contains("security-auditlog")
 
         @JvmStatic
         protected fun waitForThreadPools(client: RestClient) {
--- a/src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
+++ b/src/test/kotlin/org/opensearch/indexmanagement/IndexManagementRestTestCase.kt
@@ -364,20 +364,22 @@ abstract class IndexManagementRestTestCase : ODFERestTestCase() {
                 for ((_, value1) in nodeTasks!!) {
                     val task = value1 as Map<String, Any>
                     val action = task["action"] as String
-                    if (isIgnorableTask(action)) continue
-                    // runningTasks.add(task["action"].toString() + " | " + task["description"].toString())
+                    val description = task["description"] as? String ?: ""
+                    if (isIgnorableTask(action, description)) continue
                     runningTasks.add(task.toString())
                 }
             }
             return runningTasks
         }
 
-        // Ignore the task list API and segment replication background tasks (always running on Remote Store clusters)
-        private fun isIgnorableTask(action: String): Boolean =
+        // Ignore the task list API, segment replication background tasks, and security audit log writes
+        // (audit log writes are asynchronous and can outlive the test that triggered them)
+        private fun isIgnorableTask(action: String, description: String = ""): Boolean =
             action == ListTasksAction.NAME ||
                 action == ListTasksAction.NAME + "[n]" ||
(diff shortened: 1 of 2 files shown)


Hailong-am:
@tandonks @shiv0408 can you help to review this?

rishabh6788 (approved):

cwperks:
@DarshitChanpura @Taiwo435 can you please check how this relates to https://github.com/opensearch-project/security/pull/6341?

DarshitChanpura:
A few notes from the security-plugin side, based on the actual 3.8 code (not the PR descriptions):

- **opensearch-project/security#6341** merged to `main` ~2 days ago and is **not** on `3.8`. On 3.8 it would only add the compliance `IndexingOperationListener` (still `!disabled`-guarded there), so it isn't involved here regardless.
- **opensearch-project/security#6304** + its fast-follow **opensearch-project/security#6321** *are* on `3.8`, and to be precise they enable standalone audit in **both `ssl_only` and `disabled`** modes — not ssl_only alone. On 3.8, `createComponents` calls `initStandaloneAuditIfEnabled(...)` when `disabled`, and both the REST `AuditActionFilter` (`getActionFilters`, the `else if` branch has no `!disabled` guard) and the `AuditTransportInterceptor` (`getTransportInterceptors`, gated on `disabled || isSslOnlyMode()`) fire in disabled mode once `plugins.security.audit.type` is set.

The reason it shouldn't be affecting these runs is different: in the without-security CI here the **security plugin isn't installed at all**. The test-cluster log shows `Missing optional plugin [opensearch-security]`, and `build.gradle` only installs security under `-Dsecurity`/`-Dhttps` (it never sets `plugins.security.disabled`). With no plugin loaded, none of the standalone-audit paths execute, so `security-auditlog-*` write tasks shouldn't appear in these runs at all. Might be worth confirming whether the lingering task at teardown is actually a `security-auditlog` write vs. a background ISM / job-scheduler task.

DarshitChanpura:
Follow-up with a local reproduction, to pin down what teardown actually sees.

I ran the ISM suites that were churning at teardown (`RestRetryFailedManagedIndexActionIT`, `RolloverActionIT`, `RestExplainActionIT`) on `3.8`, **without** `-Dsecurity`/`-Dhttps` (the failing config), with the `security-auditlog` filter from this PR reverted and a diagnostic that logs every task `waitForRunningTasks` observes before filtering.

Across 710 task observations at teardown, the complete inventory was:

| Count | action | ignorable |
|------:|--------|:---------:|
| 286 | `cluster:monitor/tasks/lists` | yes (the poll itself) |
| 286 | `cluster:monitor/tasks/lists[n]` | yes |
| 76 | `indices:admin/create` | **no** |
| 62 | `internal:cluster/coordination/publish_state` | **no** |

Two takeaways:

1. **`security-auditlog` was observed 0 times.** The cluster log confirms why — the security plugin isn't installed in these runs (`Missing optional plugin [opensearch-security]`); it's only installed under `-Dsecurity`/`-Dhttps` via the `if (securityEnabled)` gate in `build.gradle`. So there is no audit logger and no `security-auditlog-*` write task to catch here.

2. The only **non-ignorable** lingering tasks are core OpenSearch `indices:admin/create` and `internal:cluster/coordination/publish_state` — async index-creation / cluster-state-publish work still settling after a test. These are the tasks that can produce a spurious "still tasks running" failure, and the `description.contains("security-auditlog")` filter wouldn't match them.

So this filter shouldn't hurt (it just never matches in without-security runs), but the actual flakiness source in these runs looks like `indices:admin/create` / `publish_state` settling, not audit logging. Happy to share the repro branch/diff if useful.

DarshitChanpura:
To reconcile this with my earlier note — the difference comes down to **two different test harnesses**, and it actually confirms the `audit.type`-gated behavior rather than contradicting it.

**1. This plugin's own GitHub Actions CI** (`test-and-build`, `multi node`, etc.)
`without-security` here means the security plugin is **not installed at all**. `build.gradle` only installs it under `-Dsecurity`/`-Dhttps` (`if (securityEnabled) { plugin(provider(securityPluginFile)) }`), and the archive distribution doesn't bundle it. The cluster log confirms it: `Missing optional plugin [opensearch-security]`. A local repro of the ISM suites in this harness showed **zero** `security-auditlog` tasks at teardown — the only non-ignorable lingering tasks were core `indices:admin/create` / `internal:cluster/coordination/publish_state`.

**2. The OpenSearch distribution integ-test** (`ci.opensearch.org`, e.g. the 3.8.0 build behind these AUTOCUT reports)
`without-security` means something different. The bundled security plugin stays **installed**, and opensearch-build just appends `plugins.security.disabled: true`:

```python
# opensearch-build src/test_workflow/integ_test/service_opensearch.py
if not self.security_enabled and os.path.isdir(self.security_plugin_dir):
    self.__add_plugin_specific_config({"plugins.security.disabled": "true"})
```

The distribution also ships `plugins.security.audit.type: internal_opensearch` (config/opensearch.yml.example + demo-config installer).

**Why that matters — and why it confirms the audit behavior:** on `3.8`, `disabled` mode + a configured `audit.type` is exactly the trigger for standalone audit logging (opensearch-project/security#6304 + its fast-follow #6321). `initStandaloneAuditIfEnabled` sees `audit.type != null` and builds a real `AuditLogImpl` (a `null` type would fall back to `NullAuditLog` and write nothing), and the REST `AuditActionFilter` runs in disabled mode. So REST calls emit `REQUEST_AUDIT` events → async bulk writes to `security-auditlog-*`, which is precisely what shows up lingering at `waitForRunningTasks` in the distribution runs.

So both observations are correct for their respective setups:
- Plugin GH Actions CI: no plugin installed → no audit tasks (the filter is inert but harmless there).
- Distribution integ-test: plugin installed + `disabled: true` + `audit.type` set → standalone audit is active by design → `security-auditlog-*` write tasks are real, and this PR's `description.contains("security-auditlog")` filter correctly skips them.

Net: the filter is the right fix for the distribution integ-test failures, and this is expected behavior of the `audit.type`-gated standalone audit path in disabled mode — not a regression.

Hailong-am:
>The OpenSearch distribution integ-test

Yes, it's coming from this. https://ci.opensearch.org/ci/dbc/integ-test/3.8.0/12102/linux/arm64/deb/test-results/11455/integ-test/index-management/without-security/opensearch-integ-test/lp7jVWb-Xw8/KTM3wysKjHo.html

PR #1712, opened 2026-08-03
Support mixed raw and rollup search when a queried field is missing from the rollup

aditi-1601:
### Description
When a search spans both raw and rollup indices and references a field that no rollup index covers, the rollup shard
previously threw `Could not find a rollup job that can answer this query because [missing field <field>]`, failing the entire search, even though the raw indices could answer it.
This change lets such mixed searches succeed by having the rollup shard contribute nothing instead of failing.
If field is present in neither the raw nor any rollup index, it succeeds with an empty aggregation result.

###Changes
  
  1. **`validateIndicies`** now tracks whether a raw index is part of the search (`hasRawIndex`). The `Could not find a rollup job...` error is thrown only when no rollup can answer and there is no raw index to fall back on. When a raw index is present, it returns an empty match instead of throwing.
  
  2. **`neutralizeRollupShard`** handles that empty case. Because the coordinator waits for every shard to return its aggregation (dropping the aggregation causes the search to hang, and running the original query against the rollup's stored structure errors on nested dimension fields such as `timestamp`), the rollup shard is instead rewritten to:
     - a `match_none` query (matches no documents), and
     - each aggregation rebuilt (neutralizeAggregation) with the same name and type but pointed at a non-existent field, so it returns an empty result the coordinator can merge with the raw shards' results.
  
The raw indices then serve the query, and the coordinator reduces the rollup shard's empty contribution together with the raw results.

### Testing
1. mixed raw + rollup search where the field exists only in the raw index:

Aggregate terms(vendor) (raw-only field) across a raw index + a rollup index. Previously failed with Could not find a rollup job that can answer this query because [missing field vendor].
  
  POST /raw_index,rollup_index/_search
  {"size":0,"aggs":{"by_vendor":{"terms":{"field":"vendor"},"aggs":{"s":{"sum":{"field":"value"}}}}}}
  
  Result: 200 OK, "_shards":{"total":2,"successful":2,"failed":0}
  buckets: acme -> 1000, books -> 3000   (identical to querying the raw index alone)

2.  date_histogram + time_zone mixed:

POST /raw_index,rollup_index/_search
  {"size":0,"aggs":{"by_day":{"date_histogram":{"field":"timestamp","calendar_interval":"1d","time_zone":"+05:30"},
   "aggs":{"v":{"terms":{"field":"vendor"}}}}}}
  
  Result: 200 OK, "_shards":{...,"failed":0}
  Day buckets + vendor sub-buckets identical to the raw-only result.

Neutralized shard preserves the query's bucketing settings (interval/time_zone/offset/bounds), so bucket keys stay aligned across shards

3. rollup-only search still errors correctly:

 When no raw index is present and no rollup covers the field, behavior is unchanged:
  
  POST /rollup-all/_search   (rollup indices only)
  {"size":0,"aggs":{"b":{"terms":{"field":"nonexistent_field"}}}}
  
  Result: 400, "Could not find a rollup job that can answer this query because [missing field nonexistent_field]"

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List

### DIFF
--- a/src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
+++ b/src/main/kotlin/org/opensearch/indexmanagement/rollup/interceptor/RollupInterceptor.kt
@@ -16,6 +16,7 @@ import org.opensearch.index.query.BoostingQueryBuilder
 import org.opensearch.index.query.ConstantScoreQueryBuilder
 import org.opensearch.index.query.DisMaxQueryBuilder
 import org.opensearch.index.query.MatchAllQueryBuilder
+import org.opensearch.index.query.MatchNoneQueryBuilder
 import org.opensearch.index.query.MatchPhraseQueryBuilder
 import org.opensearch.index.query.QueryBuilder
 import org.opensearch.index.query.QueryStringQueryBuilder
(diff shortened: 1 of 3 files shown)


tandonks:
Looks good, lets attach the testing details to the PR

tandonks (approved):

